PR #27, opened 2014-02-19
with Depends: on Rcpp (>= 0.11.0) we can remove Rscript calls...

eddelbuettel:
... from src/Makevars{,.win}

Your just-uploaded-to-CRAN version 1.2.3 has the Depends on "Rcpp (>= 0.11.0)". Which is really needed given that we no longer have a library to link against. So you may as well take advantage of this: this pull requests removes the now spurious call to Rscript to ask Rcpp about the library location and name. Because with Rcpp (>= 0.11.0) you only get an empty string back anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # Sample .travis.yml for R projects.
 #
 # See https://github.com/craigcitro/r-travis/wiki
+# for documentation, examples and more
 
 language: c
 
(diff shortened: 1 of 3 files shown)


wch:
Closing because this is no longer necessary.

eddelbuettel:
Thanks for patiently ignoring the PR for four years :)

PR #36, opened 2014-09-26
FreeBSD Build Fixes

heckendorfc:
FreeBSD names GNU make `gmake` and will require linking against libkvm.


wch:
Closing because I don't think this is necessary anymore. If there's still problems with FreeBSD, please open a new issue.

PR #44, opened 2015-03-15
Http parser 2.4.2

actsasflinn:
I've updated httpuv to use libuv version 1.4.2 + http-parser version 2.4.2


QuLogic:
This can probably be closed since #91 and #93 have done the same things.

PR #46, opened 2015-03-17
bump version to 1.3.3; require Rcpp 0.11.5

aronatkins:


kevinushey:
FWIW, Rcpp 0.12.0 no longer makes use of this warning recorder mechanism and so this PR should be unnecessary in the future.

PR #50, opened 2015-06-29
Don't add Content-Length header if already present

wch:
This is to support rstudio/shiny#876.

The `RcppExports.cpp` file had a lot of changes - perhaps because I built this with a newer version of Rcpp (0.11.6)?

### DIFF
--- a/src/http.cpp
+++ b/src/http.cpp
@@ -392,13 +392,21 @@ void HttpResponse::writeResponse() {
   // TODO: Optimize
   std::ostringstream response(std::ios_base::binary);
   response << "HTTP/1.1 " << _statusCode << " " << _status << "\r\n";
+  bool hasContentLengthHeader = false;
   for (ResponseHeaders::const_iterator it = _headers.begin();
      it != _headers.end();
      it++) {
     response << it->first << ": " << it->second << "\r\n";
+
+    if (it->first == "Content-Length") {
+      hasContentLengthHeader = true;
+    }
   }
-  if (_pBody)
+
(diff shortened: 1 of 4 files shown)


src/http.cpp
```diff
@@ -392,13 +392,21 @@ void HttpResponse::writeResponse() {
   // TODO: Optimize
   std::ostringstream response(std::ios_base::binary);
   response << "HTTP/1.1 " << _statusCode << " " << _status << "\r\n";
+  bool hasContentLengthHeader = false;
   for (ResponseHeaders::const_iterator it = _headers.begin();
      it != _headers.end();
      it++) {
     response << it->first << ": " << it->second << "\r\n";
+
+    if (it->first == "Content-Length") {
```
aronatkins: HTTP headers are case insensitive. For example, I've seen some pretty wacky bugs crop up because A was using `Content-Type` while B was expecting `Content-type`. You should perform this comparison in a case insensitive manner.

aronatkins: There's another `Content-Length` reference in `HttpRequest::_on_headers_complete` ..

aronatkins: Sorry; those headers are safe because `RequestHeaders` use case-insensitive comparisons.

wch: Thanks, fixed!

jjallaire:
Yes, the RcppExports.cpp changes are a result of the new version and they are nothing to worry (the new generated code is functionally equivalent to the old code but is just syntactically a bit simpler).

PR #55, opened 2015-10-13
update Makevars

tljm:
In OpenBSD (as in FreeBSD) -lkvm is required to compile httmpuv.


wch:
Closing because we've already incorporated this change.

PR #57, opened 2015-11-18
Allow creating a TCP server using an ipv6 address

onurfiliz:
For ipv6-only setups, we need to be able to pass ipv6 addresses as the
host (e.g. '::') argument to httpuv::startServer().

Not sure if this is the right approach, I'd love to get your feedback.


onurfiliz:
Superseded by #115.

PR #58, opened 2016-01-30
Complete Win32 non-blocking implementation

hcorrada:
Using ideas from Simon Urbanek's 'background' draft package that implements the Rhttpd strategy. Implementation description in `src/httpuv.cpp`


src/httpuv.cpp
```diff
@@ -486,11 +487,11 @@ class DaemonizedServer {
...
     }
     #else 
-      if (server_thread) {
+      if (thread) {
         DWORD ts = 0;
-        if (GetExitCodeThread(server_thread, &ts) && ts == STILL_ACTIVE)
-          TerminateThread(server_thread, 0);
-        server_thread = 0;
+        if (GetExitCodeThread(thread, &ts) && ts == STILL_ACTIVE)
+          TerminateThread(thread, 0);
```
jcheng5: This seems really bad--it's almost never a good idea to forcibly terminate a thread from the outside, with the possible exception of when the entire process is about to shut down. It can leave all sorts of state corrupted. The usual approach is to set a flag that the other thread will check, and if it's important for this thread to not continue until it knows the other thread is dead, to wait/join the other thread (in Windows it looks like `WaitForSingleObject` will work).

jcheng5:
Thanks for all the hard work. I know this stuff takes a lot of effort.

I don't fully understand the comment in httpuv.cpp; is the libuv loop run on the main thread, or the background thread?

jcheng5:
I may also be spending more time on this in the future. It turns out that even in the non-daemonized case, having the libuv loop running on the main thread really hurts I/O performance a lot when R is also busy--it's enough of a problem that some geospatial related features in Leaflet are not worth shipping until we resolve it.

src/httpuv.cpp
hcorrada: Thanks. I'll take a look, these lines are straight from 'background' and Rhttpd, so I'll pass along whatever we end up using here to them.

hcorrada:
The libuv ends up running in the main thread, through this message passing mechanism. The background thread is only running the 'polling' loop. Again, straight from 'background' and Rhttpd.

What do you have in mind? With the existing 'InputHandler' and this win32 message passing mechanisms, the libuv loop still runs in the main R thread. Perhaps use libuv's thread mechanism directly?

Ok, thanks, I'll add a better solution to thread stopping in the next day or so.

jjallaire:
If the background thread is running a message loop then you should just be
able to "PostMessage" (async message pass) a custom WM_ message telling it
to terminate (you just then handle that message within the thread by call
ExitThread.

hcorrada:
@jcheng5 Check this thread exit strategy. Let me know if it can be improved. THanks!

src/httpuv.cpp
```diff
@@ -439,43 +444,73 @@ std::string base64encode(const Rcpp::RawVector& x) {
...
+  void stop_thread() {
+    DBG(Rprintf("called stop_thread\n"));
+    if (!thread) return;
+    
+    // signal thread exit
+    DWORD ts = 0;
+    int ntimes = 10;
+    int i = 0;
+    
+    while (i < ntimes && GetExitCodeThread(thread, &ts) && ts == STILL_ACTIVE) {
```
jcheng5: I think you just want `WaitForSingleObject(hThread, millis)` here.

src/httpuv.cpp
```diff
@@ -486,42 +521,160 @@ class DaemonizedServer {
...
+{
+  DaemonizedServer *dServer = (DaemonizedServer *) lpParameter;
+  if (!dServer) return 0;
+  
+  // run the libuv loop as long as thread exit hasn't been triggered
+  while ((dServer->thread_status & THREAD_DISPOSE) == 0) {
+    DBG(Rprintf("calling run_libuv from LibuvThreadProc\n"));
+    run_libuv(dServer);
+    DBG(Rprintf("done run_libuv from LibuvThreadProc\n"));
+    Sleep(1);
```
jcheng5: Does this introduce up to 1 second of latency?

hcorrada:
With `WaitForSingleObject` I can take that out. I'll submit another commit soon.

hcorrada:
Hey @jcheng5, thanks for the pointer!

hcorrada:
Hi, @jcheng5, can this be pulled in?

paul-shannon:
Hi Hector, 

 @hcorrada  - I cannot tell if your good work, getting httpuv async communications, R <-> browser, made it into release.   Could you let me know?

hcorrada:
Don't think so... @jcheng5 maybe take a look after rstudio conference?

paul-shannon:
Thanks, Hector.  Here’s hoping!  

 - Paul

jcheng5:
Sorry for the long, long delay on this. I just felt really uncomfortable with the Windows implementation using threads and still do. I believe I may have found a much cleaner solution using SetTimer though, this doesn't need a background thread at all:
https://github.com/jcheng5/later/blob/master/src/later_win32.cpp

I also need to dive deeply back into httpuv to support asynchronous I/O. I'll try to prioritize Win32 support for daemonized httpuv as well, and getting it to a place where I'm comfortable calling it non-experimental.

paul-shannon:
Hi Joe,

This is great news!   Daemonized httpuv on all platforms would be a real boon.

 - Paul

hcorrada:
Thanks Joe! That would be awesome!

wch:
I think this is no longer necessary because of #106.

Issue #63, opened 2016-09-11
Installation fails on the Alpine Linux


artemklevtsov:
Hi.

I can't build this package on Alpine Linux.

To reproduce:

``` bash
docker run -ti --rm alpine:edge ash
apk update
apk add R R-dev g++
echo 'options(repos = c(CRAN = "https://cloud.r-project.org/"))' >> /usr/lib/R/etc/Rprofile.site
Rscript -e 'install.packages("httpuv")'
```

The same errors with git-version.

Build log:

``` bash
also installing the dependency ‘Rcpp’

trying URL 'https://cloud.r-project.org/src/contrib/Rcpp_0.12.7.tar.gz'
Content type 'application/x-gzip' length 2421965 bytes (2.3 MB)
==================================================
downloaded 2.3 MB

trying URL 'https://cloud.r-project.org/src/contrib/httpuv_1.3.3.tar.gz'
Content type 'application/x-gzip' length 426665 bytes (416 KB)
==================================================
downloaded 416 KB

* installing *source* package ‘Rcpp’ ...
** package ‘Rcpp’ successfully unpacked and MD5 sums checked
** libs
g++ -I/usr/include/R -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer -D__MUSL__  -c Date.cpp -o Date.o
g++ -I/usr/include/R -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer -D__MUSL__  -c Module.cpp -o Module.o
g++ -I/usr/include/R -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer -D__MUSL__  -c Rcpp_init.cpp -o Rcpp_init.o
g++ -I/usr/include/R -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer -D__MUSL__  -c api.cpp -o api.o
g++ -I/usr/include/R -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer -D__MUSL__  -c attributes.cpp -o attributes.o
[...]

PR #64, opened 2016-09-11
Fix build error with musl (#63)

artemklevtsov:
Based on the Alpine APKBUILD: http://git.alpinelinux.org/cgit/aports/tree/main/libuv/APKBUILD#n34


wch:
I think this is no longer needed -- AFAIK we've been able to build on Alpine for a while now. If that's not the case, please open an issue with a reproducible example. For example, this comment on an issue provides a Dockerfile that allowed us to reproduce the problem and fix it: https://github.com/r-lib/later/issues/45#issuecomment-411584635

PR #66, opened 2016-12-04
Gifi patch 1

### LINKED ISSUE #47: Install fails when .Rprofile prints to console
I'm using `install.packages("httpuv", type="source")` to install on a machine that uses a .Rprofile or Rprofile.site file that prints to the console:

```
.First <- function() {
  cat("Custom .Rprofile")
}
```

The build fails with a `No such file or directory` message, and the problem line is

```
g++ -m32 -shared -s -static-libgcc -o httpuv.dll tmp.def RcppExports.o filedatasource-unix.o filedatasource-win.o http.o httpuv.o md5.o set_invalid_parameter.o uvutil.o websockets-base.o websockets-hixie76.o websockets-hybi03.o websockets-ietf.o websockets.o Custom .Rprofile ./libuv/libuv.a ./http-parser/http_parser.o ./sha1/sha1.o ./base64/base64.o -lWs2_32 -lkernel32 -lpsapi -liphlpapi -Ld:/RCompile/r-compiling/local/local320/lib/i386 -Ld:/RCompile/r-compiling/local/local320/lib -LC:/PROGRA~1/R/R-32~1.0/bin/i386 -lR
g++.exe: error: Custom: No such file or directory
g++.exe: error: .Rprofile: No such file or directory
```

It appears that the `cat` output from the .Rprofile on this machine is being inserted into the g++ command.  After removing the `cat` call, the library installs without error.  

I've been able to reproduce this error on both windows and linux machines.  This may also be related to #11.

gifi:
When installing from sources, the PR fixes errors like
```
g++: error: >: No such file or directory
g++: error: Rcpp:::LdFlags(): No such file or directory
g++: error: >: No such file or directory
g++: error: >: No such file or directory
```
when an R startup script has ```cat()```.

This will fix #47.

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -6,7 +6,7 @@ ifeq ($(UNAME), Darwin)
 FRAMEWORK = -framework CoreServices
 endif
 
-PKG_LIBS = `$(R_HOME)/bin/Rscript -e "Rcpp:::LdFlags()"` ./libuv/libuv.a ./http-parser/http_parser.o ./sha1/sha1.o ./base64/base64.o $(FRAMEWORK)
+PKG_LIBS = ./libuv/libuv.a ./http-parser/http_parser.o ./sha1/sha1.o ./base64/base64.o $(FRAMEWORK)
 ifeq ($(UNAME), SunOS)
 PKG_LIBS += -lkstat -lsendfile
 endif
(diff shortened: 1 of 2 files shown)


yihui:
I think these changes are long overdue (also see #27) and we should accept this PR. cc @jcheng5

PR #67, opened 2017-04-10
Support httpuv::service(NA) for non-blocking run

jeroen:
This adds support for performing a non-blocking run `UV_RUN_NOWAIT ` via:

```r
httpuv::service(NA)
```
The implementation is trivial, but unfortunately `httpuv::service(0)` was already taken to mean "no timeout" instead of "zero timeout". Therefore this uses `NA` for this purpose.

### DIFF
--- a/R/httpuv.R
+++ b/R/httpuv.R
@@ -397,7 +397,7 @@ startPipeServer <- function(name, mask, app) {
 #' 
 #' @param timeoutMs Approximate number of milliseconds to run before returning. 
 #'   If 0, then the function will continually process requests without returning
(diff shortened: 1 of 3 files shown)


jcheng5:
Thanks for this. Have you ever submitted a CLA for RStudio? If so, I can merge this right away.

jeroen:
Yes, I've done contracting work for hadley and JJ.

jcheng5:
Thanks!

jeroen:
I forgot to bump the version. Can you maybe increment so that I can use this feature conditionally? Thanks!

PR #73, opened 2017-06-27
Add getRNGState function

wch:
This is needed for rstudio/shiny#1763.

Note: there's some bug in the current version of roxygen that makes it error when I press Cmd-Shift-D in RStudio, or run `roxygen2::roxygenise()`. Running `devtools::document()` works though. The error appears to be related to reference classes. Another reason to move to R6... :)


jcheng5:
Maybe `@keywords internal`? Otherwise, LGTM.

PR #80, opened 2017-09-12
Implement async

wch:
This converts `AppWrapper$onHeaders` and `AppWrapper$call` to async. It requires the async branch of rstudio/shiny.

One issue that still needs to be dealt with is how to handle errors in the header. In the old synchronous version of `HttpRequest::_on_headers_complete`, if there was an error, it returned 2, and the http parser would [set an error flag and return](https://github.com/rstudio/httpuv/blob/65016f205bdfca17d7714a44740682272bd1c25c/src/http-parser/http_parser.c#L1570-L1592). Now, we can't know that there's an error until later
but now we're not able to tell there's an error until the `_on_headers_complete_complete` callback is invoked, so we can't tell the http parser to set the error condition and return early.

It might also be nice to wrap up the C++ callback into an R function so that in R, instead of calling `invoke_cpp_callback(., cpp_callback)` (where `cpp_callback` is an external pointer to a C++ function), it can simply call `callback(.)`.


jcheng5:
I'm so happy to see this!

I don't have an answer to the on_headers_complete thing off the top of my head. I don't remember, do we have the option of not supporting async for anything but app$call()?

If you have time, can you try benchmarking with e.g. Apachebench and see if we lost significant perf vs what's on master? (Sync on master vs. Sync on branch vs. Async on branch) And have you done any testing for memory leaks?

wch:
Running with valgrind, there doesn't seem to be any (new) memory leaks introduced by this.

To test, I ran a simple Shiny app with `R -d valgrind`, opening in a browser, and reloading once.

Results from CRAN versions of Shiny (1.0.5) and httpuv (1.3.5):

```
==26921== 
==26921== HEAP SUMMARY:
==26921==     in use at exit: 61,507,015 bytes in 35,684 blocks
==26921==   total heap usage: 125,987 allocs, 90,303 frees, 185,022,573 bytes allocated
==26921== 
==26921== LEAK SUMMARY:
==26921==    definitely lost: 8,016 bytes in 30 blocks
==26921==    indirectly lost: 9,855 bytes in 434 blocks
==26921==      possibly lost: 1,524 bytes in 21 blocks
==26921==    still reachable: 61,486,676 bytes in 35,193 blocks
==26921==                       of which reachable via heuristic:
==26921==                         length64           : 80 bytes in 2 blocks
==26921==                         newarray           : 1,568 bytes in 18 blocks
==26921==         suppressed: 0 bytes in 0 blocks
==26921== Rerun with --leak-check=full to see details of leaked memory
==26921== 
==26921== For counts of detected and suppressed errors, rerun with: -v
==26921== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

Using the async branch of Shiny and this branch of httpuv:

```
==26738== 
==26738== HEAP SUMMARY:
==26738==     in use at exit: 72,140,058 bytes in 41,307 blocks
==26738==   total heap usage: 135,194 allocs, 93,887 frees, 197,130,858 bytes allocated
==26738== 
==26738== LEAK SUMMARY:
==26738==    definitely lost: 8,016 bytes in 30 blocks
==26738==    indirectly lost: 9,855 bytes in 434 blocks
==26738==      possibly lost: 1,828 bytes in 22 blocks
==26738==    still reachable: 72,119,415 bytes in 40,815 blocks
==26738==                       of which reachable via heuristic:
==26738==                         length64           : 80 bytes in 2 blocks
==26738==                         newarray           : 1,568 bytes in 18 blocks
==26738==         suppressed: 0 bytes in 0 blocks
==26738== Rerun with --leak-check=full to see details of leaked memory
==26738== 
==26738== For counts of detected and suppressed errors, rerun with: -v
==26738== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

(Note that the results vary from run to run.)

wch:
I currently get a segfault while trying to debug some of Shiny's middleware. 

```
# Start in shiny directory, on async branch
devtools::load_all()
debug(handlerManager$handlers$invoke)
shinyApp(
  ui = fluidPage(
    numericInput("n", "n", 1),
    plotOutput("plot")
  ),
  server = function(input, output) {
    output$plot <- renderPlot( plot(head(cars, input$n)) )
  }
)

# At browser prompt, type Q

#  *** caught segfault ***
# address 0x0, cause 'unknown'
# 
# Traceback:
#  1: .Call("httpuv_invokeCppCallback", PACKAGE = "httpuv", data, callback_sexp)
#  2: invokeCppCallback(., cpp_callback)
#  3: onFulfilled(value, visible)
#  4: withVisible(value)
#  5: private$doResolve(value)
#  6: doTryCatch(return(expr), name, parentenv, handler)
#  7: tryCatchOne(expr, names, parentenv, handlers[[1L]])
#  8: tryCatchList(expr, classes, parentenv, handlers)
#  9: tryCatch({    private$doResolve(value)}, error = function(err) {    private$doReject(err)})
# 10: resolve(onFulfilled(value, visible))
# 11: f(private$value, private$visible)
# 12: FUN(X[[i]], ...)
# 13: lapply(private$onFulfilled, function(f) {    f(private$value, private$visible)})
# 14: (function () {    lapply(private$onFulfilled, function(f) {        f(private$value, private$visible)    })    private$onFulfilled <- list()})()
# 15: evalq((function () {    lapply(private$onFulfilled, function(f) {        f(private$value, private$visible)    })    private$onFulfilled <- list()})(), <environment>)
# 16: evalq((function () {    lapply(private$onFulfilled, function(f) {        f(private$value, private$visible)    })    private$onFulfilled <- list()})(), <environment>)
# 17: doTryCatch(return(expr), name, parentenv, handler)
# 18: tryCatchOne(expr, names, parentenv, handlers[[1L]])
# 19: tryCatchList(expr, names[-nh], parentenv, handlers[-nh])
# 20: doTryCatch(return(expr), name, parentenv, handler)
# 21: tryCatchOne(tryCatchList(expr, names[-nh], parentenv, handlers[-nh]),     names[nh], parentenv, handlers[[nh]])
# 22: tryCatchList(expr, classes, parentenv, handlers)
# 23: tryCatch(evalq((function () {    lapply(private$onFulfilled, function(f) {        f(private$value, private$visible)    })    private$onFulfilled <- list()})(), <environment>), error = function (x) x, interrupt = function (x) x)
```

Continuing with `c` works; it's quitting with `Q` that causes the problem.

wch:
Perhaps the promise code in `rookCall` should be restructured so that the `func` is not run immediately -- that is, so that instead of being in `promise()`, it's in a `then()`.

wch:
After looking at the code some more, I think the onheaders part needs to be synchronous. Otherwise, this could happen: the `onHeaders` and `call` callbacks are both queued, then the `onHeaders` callback reports an error. This should prevent the `call` callback from executing, but in this the `call` callback will still execute.

wch:
For an extremely simple app (see below), it looks like there is a slight performance hit for this change. For 10 concurrent connections with 200 total, with the old version it takes about 1.0 seconds, and with the async version it takes about 1.2 seconds. The average time pre request is 50ms vs 57ms.

This is the app:

```R
library(httpuv)
app <- list(
  call = function(req) {
    list(
      status = 200L,
      headers = list(
        'Content-Type' = 'text/html',
        Connection = 'keep-alive'
      ),
      body = "abc"
    )
  } 
)

runServer("0.0.0.0", 5000, app)
```

The results from httpuv 1.3.5:

```
$ docker run --rm jordi/ab ab -k -n 500 -c 100  http://10.0.0.53:5000/foo.html
...
Concurrency Level:      10
Time taken for tests:   0.998 seconds
Complete requests:      200
Failed requests:        0
Keep-Alive requests:    200
Total transferred:      18000 bytes
HTML transferred:       600 bytes
Requests per second:    200.47 [#/sec] (mean)
Time per request:       49.882 [ms] (mean)
Time per request:       4.988 [ms] (mean, across all concurrent requests)
Transfer rate:          17.62 [Kbytes/sec] received
...
```

And from this branch:

```
$ docker run --rm jordi/ab ab -k -n 500 -c 100  http://10.0.0.53:7000/foo.html
...
Concurrency Level:      10
Time taken for tests:   1.157 seconds
Complete requests:      200
Failed requests:        0
Keep-Alive requests:    200
Total transferred:      18000 bytes
HTML transferred:       600 bytes
Requests per second:    172.85 [#/sec] (mean)
Time per request:       57.855 [ms] (mean)
Time per request:       5.785 [ms] (mean, across all concurrent requests)
Transfer rate:          15.19 [Kbytes/sec] received
...
```

*****

With more concurrent connections, the async branch seems to suffer more. Using 100 concurrent, 500 total, the CRAN version took 74ms per request, while the async version took  250ms per request.

httpuv 1.3.5:

```
Concurrency Level:      100
Time taken for tests:   0.370 seconds
Complete requests:      500
Failed requests:        0
Keep-Alive requests:    500
Total transferred:      45000 bytes
HTML transferred:       1500 bytes
Requests per second:    1352.10 [#/sec] (mean)
Time per request:       73.959 [ms] (mean)
Time per request:       0.740 [ms] (mean, across all concurrent requests)
Transfer rate:          118.84 [Kbytes/sec] received
```

This branch:

```
Concurrency Level:      100
Time taken for tests:   1.246 seconds
Complete requests:      500
Failed requests:        0
Keep-Alive requests:    500
Total transferred:      45000 bytes
HTML transferred:       1500 bytes
Requests per second:    401.44 [#/sec] (mean)
Time per request:       249.102 [ms] (mean)
Time per request:       2.491 [ms] (mean, across all concurrent requests)
Transfer rate:          35.28 [Kbytes/sec] received
```

wch:
There's now a fast path if the user has provides a non-async handler (the common case), and the slower path using promises if the user provides an async handler.

Here's a test app:

```R
library(promises)
library(httpuv)

app <- list(
  call = function(req) {
    content <- list(
      status = 200L,
      headers = list(
        'Content-Type' = 'text/html',
        Connection = 'keep-alive'
      ),
      body = "abc"
    )
    
    if (req$PATH_INFO %in% c("/", "/sync")) {
      return(content)

    } else if (req$PATH_INFO == "/sync-error") {
      stop("Error in app (sync)")

    } else if (req$PATH_INFO == "/async") {
      promise(function(resolve, reject) {
        resolve(content)
      })

    } else if (req$PATH_INFO == "/async-error") {
      promise(function(resolve, reject) {
        stop("Error in app (async)")
      })

    } else {
      stop("Unknown request path:", req$PATH_INFO)
    }
  } 
)

runServer("0.0.0.0", 5000, app)
```

Note that `ab` doesn't work correctly with the error URLs because the 500 error page provided by httpuv doesn't have the keep-alive header that `ab` needs. (This is due to httpuv's incomplete support of HTTP 1.0.)

Results for sync:

```
$ ab -k -n 1000 -c 100 http://127.0.0.1:5000/sync
...
Concurrency Level:      100
Time taken for tests:   0.426 seconds
Complete requests:      1000
Failed requests:        0
Keep-Alive requests:    1000
Total transferred:      90000 bytes
HTML transferred:       3000 bytes
Requests per second:    2348.70 [#/sec] (mean)
Time per request:       42.577 [ms] (mean)
Time per request:       0.426 [ms] (mean, across all concurrent requests)
Transfer rate:          206.43 [Kbytes/sec] received
```

Results for async:

```
$ ab -k -n 1000 -c 100 http://127.0.0.1:5000/async
...
Concurrency Level:      100
Time taken for tests:   2.211 seconds
Complete requests:      1000
Failed requests:        0
Keep-Alive requests:    1000
Total transferred:      90000 bytes
HTML transferred:       3000 bytes
Requests per second:    452.23 [#/sec] (mean)
Time per request:       221.125 [ms] (mean)
Time per request:       2.211 [ms] (mean, across all concurrent requests)
Transfer rate:          39.75 [Kbytes/sec] received
```

The performance for synchronous handlers is about 2x of the previous test, because of the caching of the package version.

R/httpuv.R
```diff
@@ -410,6 +452,7 @@ startPipeServer <- function(name, mask, app) {
 #' @export
 service <- function(timeoutMs = ifelse(interactive(), 100, 1000)) {
   run(timeoutMs)
+  later::run_now()
```
jcheng5: What do you think of making a habit out of documenting functions that will, directly or indirectly, call `run_now`? If you're writing promises in some script and then call a synchronous function that calls `later::run_now`, it could change your assumptions about the order things happen in.

```r
future(a()) %...>% b()
c()
d()
```

Normally, you'd expect that `c()` and `d()` both execute before `b()`. But if `c()` indirectly calls `later::run_now`--e.g. runs a Shiny gadget--then `b()` will execute before `d()`.

wch: Yeah, that's a good idea. `run_now()` has side-effects that can potentially alter global state, so it would be good for developers to make it clear when they're using it.

This brings to mind a couple ideas:
* Would it make sense to add some sort of namespace or scope to `later()` callbacks so that a `run_now()` could just run callbacks from a particular scope? I'm thinking something like reactive domains. Then you could call `run_now()` and not worry about executing someone else's callbacks. This might not be a common enough situation to warrant the added complexity, though.
* If someone put a `browser()` in  `b`, would the `run_now()` show in the stack trace?

jcheng5: We are planning to implement something like that for Gabor. But it is only generally useful if you can ensure that the computations outside and inside your function don't need to overlap. That might be hard for something as general as httpuv/shiny: for example, passing a promise into a function that launches a gadget. 

I would expect b() to be on the stack, yes.

jcheng5: I mean run_now() to be on b() stack.

PR #85, opened 2017-09-21
Use Rcpp symbol registration and visibility

jeroen:
Use the new automatic Rcpp symbol registration. Also compile with `-fvisibility=hidden` on linux to prevent conflicts with other packages using the same names.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: httpuv
 Type: Package
 Title: HTTP and WebSocket Server Library
-Version: 1.3.5
+Version: 1.3.5.9000
 Author: Joe Cheng, Hector Corrada Bravo [ctb], Jeroen Ooms [ctb],
     Winston Chang [ctb]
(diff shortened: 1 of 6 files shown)


wch:
@jeroen, do you know if it's possible to manually register one more function, with a specific name? The one I'm thinking of is in this file:
https://github.com/rstudio/httpuv/pull/83/files#diff-c52527bad50133555ea298b79bd86ced

It has the name `httpuv_decodeURIComponent` instead of the new `_ httpuv_decodeURIComponent`, for backward compatibility.

If this function isn't available, then all binary builds of Shiny that were built with httpuv <=1.3.5 will break.

jeroen:
Hmmm that's unfortunately. You'd need to ask @jjallaire but I don't think Rcpp currently has a way to add custom symbols to the registration template.

jjallaire:
It will be picked up and registered so long as it's recognized by  `tools::package_native_routine_registration_skeleton` (we call that function in addition to our own processing to pickup additional symbols). I believe that if you reference the symbol in a `.Call` somewhere within your R source then it will be recognized.

jeroen:
Thanks! Added this to the PR and it seems to work.

PR #91, opened 2017-10-11
Upgrade to libuv 1.15.0

wch:
This closes #42. I've tested it with `shiny::runExample("01_hello")` with Shiny 1.0.5 on macOS 10.12.6, Windows 10, and Centos 6.

Unlike the libuv-1.8.0 branch, on Windows, this builds libuv using Rtools, by simply running `make -f Makefile.mingw`. (Thanks to @jimhester!)

* The first commit just updates the files for libuv
* The second commit makes modifications to httpuv code so that it all works. For the second commit, many changes were borrowed from 5a4ba77.

PR #93, opened 2017-10-18
Upgrade to http-parser 2.7.1

wch:


QuLogic:
Same question as #91? (I'm wondering if httpuv could be built against a system copy now).

wch:
Our copy of http-parser has been modified from the original sources, so building against a system copy of it would not work.

QuLogic:
Thanks! Is there any chance these changes might go upstream?

wch:
Nope. They're very customized for the specific needs of httpuv.

PR #97, opened 2017-10-25
Call onHeaders asynchronously with later()

### LINKED ISSUE #96: Incorrect behavior with onHeaders
As of httpuv 1.3.5, the `onHeaders()` function does some things incorrectly.

With this test app:

```R
runServer("0.0.0.0", 5000,
  list(
    onHeaders = function(req) {
      list(
        status = 200L,
        body = "abc"
      )
    } 
  )
)
```

Then do a `GET /`:

```
$ nc localhost 5000
GET /

HTTP/1.1 200 OK
Content-Type: text/html
Content-Length: 3
```

Two things happen incorrectly (I think):

1. It doesn't return the body ("abc").
2. In the R side, it says there's a parse error, as in `ERROR: [on_request_read] parse error`.

I just tested with Shiny, and indeed it doesn't send the body [here](https://github.com/rstudio/shiny/blob/ba8d79f/R/middleware.R#L303).

After the server sends the data, it closes the connection. This differs from the behavior when `$call()` is used, where the server does not close the connection. Is the disconnection expected?

```
> sessioninfo::session_info()
─ Session info ────────────────────────────────────────────────────────────────────
 setting  value                       
 version  R version 3.4.1 (2017-06-30)
 os       macOS Sierra 10.12.6        
 system   x86_64, darwin15.6.0        
 ui       RStudio                     
 language (EN)                        
 collate  en_US.UTF-8                 
 tz       America/Chicago             
 date     2017-10-24                  

─ Packages ────────────────────────────────────────────────────────────────────────
 package     * version date       source         
 clisymbols    1.2.0   2017-05-21 cran (@1.2.0)  
 httpuv      * 1.3.5   2017-07-04 CRAN (R 3.4.1) 
 Rcpp          0.12.13 2017-09-28 cran (@0.12.13)
 sessioninfo   1.0.0   2017-06-21 CRAN (R 3.4.1) 
 withr         2.0.0   2017-07-28 cran (@2.0.0)  
 yaml          2.1.14  2016-11-12 CRAN (R 3.4.0) 
```

wch:
This pull request uses `later()` to call the user's `onHeaders()` function asynchronously. This is a necessary step for running the I/O on a background thread, but I thought it deserved its own PR and code review.

This also fixes #96.

### DIFF
--- a/src/http-parser/http_parser.c
+++ b/src/http-parser/http_parser.c
@@ -344,6 +344,8 @@ enum state
   /* When the (async) on_headers function has been called but we're still
    * waiting for a response. */
   , s_wait_for_on_headers_completed
+  /* After the (async) on_headers function has finished processing, and has
+   * set a result value. */
   , s_on_headers_completed
 
   , s_headers_done
@@ -1818,10 +1820,7 @@ reexecute:
            parser->method == HTTP_CONNECT);
 
         if (settings->on_headers_complete) {
-          /* Update the internal and externally-visible state. */
           UPDATE_STATE(s_wait_for_on_headers_completed);
-          parser->waiting_for_headers_completed = 1;
-
 
           /* If on_headers_complete is async, then we'll let the external code
            * set the result in parser->headers_status, and advance the state to
@@ -2151,13 +2150,19 @@ error:
   RETURN(p - data);
 }
 
-void http_parser_on_headers_completed(http_parser *parser) {
-  /* We unfortunately have to set the state in two places. parser->state is
-   * used internally by the parser, and parser->waiting_for_headers_completed
-   * externally visible.
-   */
+int http_parser_waiting_for_headers_completed(http_parser *parser) {
+  return parser->state == s_wait_for_on_headers_completed;
+}
+
+void http_parser_headers_completed(http_parser *parser, int status) {
+  if (parser->is_running) {
+    /* If this function is being called when http_parser_execute is still on
+     * the call stack, something has gone wrong.
+     */
+    assert(0 && "Called http_parser_set_headers_result inside of http_parser_execute");
+  }
+  parser->headers_status = status;
   parser->state = s_on_headers_completed;
-  parser->waiting_for_headers_completed = 0;
 }
 
 /* Does the parser need to see an EOF to find the end of the message? */
@@ -2221,7 +2226,6 @@ http_parser_init (http_parser *parser, enum http_parser_type t)
   parser->http_errno = HPE_OK;
 
   parser->is_running = 0;
-  parser->waiting_for_headers_completed = 0;
   parser->headers_status = -1;
 }
 
--- a/src/http-parser/http_parser.c
+++ b/src/http-parser/http_parser.c
@@ -60,6 +60,7 @@ do {                                                                 \
 #define UPDATE_STATE(V) p_state = (enum state) (V);
 #define RETURN(V)                                                    \
 do {                                                                 \
+  parser->is_running = 0;                                            \
   parser->state = CURRENT_STATE();                                   \
   return (V);                                                        \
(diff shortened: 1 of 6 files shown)


src/http-parser/http_parser.c
```diff
@@ -2090,6 +2151,14 @@ size_t http_parser_execute (http_parser *parser,
   RETURN(p - data);
 }
 
+void http_parser_on_headers_completed(http_parser *parser) {
```
wch: Should check `is_running` here and error if true.

src/http-parser/http_parser.c
```diff
@@ -2150,12 +2219,18 @@ http_parser_init (http_parser *parser, enum http_parser_type t)
   parser->type = t;
   parser->state = (t == HTTP_REQUEST ? s_start_req : (t == HTTP_RESPONSE ? s_start_res : s_start_req_or_res));
   parser->http_errno = HPE_OK;
+
+  parser->is_running = 0;
+  parser->waiting_for_headers_completed = 0;
```
wch: Might be better to have a function that returns true if state is `s_wait_for_on_headers_completed`.

src/http-parser/http_parser.c
```diff
@@ -2150,12 +2219,18 @@ http_parser_init (http_parser *parser, enum http_parser_type t)
...
+  parser->headers_status = -1;
 }
 
 void
 http_parser_settings_init(http_parser_settings *settings)
 {
   memset(settings, 0, sizeof(*settings));
+  /* Default is to expect synchronous on_headers_complete. */
+  settings->is_async_on_headers_complete = 0;
```
wch: Not necessary since the memory is zeroed out. Same with above.

src/http.cpp
```diff
@@ -211,48 +233,85 @@ int HttpRequest::_on_header_value(http_parser* pParser, const char* pAt, size_t
...
+  this->_pWebApplication->onHeaders(
+    this,
+    boost::bind(&HttpRequest::_on_headers_complete_complete, this, _1)
+  );
+}
+
+// This wrapper function is needed to use HttpRequest::_call_r_on_headers with
+// later(). That method can't be passed to later(), but this function can.
+void call_r_on_headers_wrapper(void* data) {
+  HttpRequest* req = (HttpRequest*) data;
```
jcheng5: Use a C++ style cast

src/http.cpp
```diff
@@ -211,48 +233,85 @@ int HttpRequest::_on_header_value(http_parser* pParser, const char* pAt, size_t
...
     }
   }
 
-  // TODO: Allocate body
-  return result;
+  // Tell the parser what the result was.
+  this->_parser.headers_status = result;
+
+  // Tell the http parser it can move on.
+  http_parser_on_headers_completed(&(this->_parser));
```
wch: Combine this and previous line into a single function call.

Rename function to `http_parser_set_headers_result`, or something like that.

src/http.cpp
```diff
@@ -303,50 +362,74 @@ void HttpRequest::close() {
   uv_close(toHandle(&_handle.stream), HttpRequest_on_closed);
 }
 
+void HttpRequest::_parse_http_data(char* buffer, const ssize_t n) {
+  int parsed = http_parser_execute(&_parser, &request_settings(), buffer, n);
+
+  if (_parser.waiting_for_headers_completed) {
+    // If we're waiting for the header response, just store the data in the
+    // buffer.
+    _requestBuffer.append(buffer + parsed, n - parsed);
```
wch: Change `_requestBuffer` to a `vector<char>`.

wch:
@jcheng5 I've addressed all the issues we discussed in the review, with one exception: I left the explicit initialization to 0 for clarity.

Issue #72, opened 2017-06-12
Missing option for closing connection after sending response.


zimnyx:
Hi! By default, connection is not closed on server side during normal processing, you can see it with simples endpoint:
grappa@adel /tmp/httpuv/demo $ curl -v http://127.0.0.1:9876
* Rebuilt URL to: http://127.0.0.1:9876/
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 9876 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:9876
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/html
< Content-Length: 3
< 
* Connection #0 to host 127.0.0.1 left intact
(see last line). 

There is no option to force closing connection at the end of response. It could be done by adding "Connection: close" header support in case of HTTP/1.1

jcheng5:
Good idea. As a workaround, you can add the header to the response yourself.

Just curious, what are you using httpuv for?

zimnyx:
Hi Joe,
thanks for interest!
I renamed this ticket, because it may be not related to disconnecting issue.

Following your suggestion, I just tried sending "Connection: close" header from server, but it didn't fixed issue with both ab and /usr/lib/nagios/plugins/check_http. These are standard tools very widely used.

So here's what I observed (httpuv 1.3.3):


Here's the endpoint:
```
library(httpuv)

app <- list(
  call = function(req) {
    list(
      status = 200L,
      headers = list(
        'Content-Type' = 'text/html',
        'Connection' = 'close'
      ),
      body = "foo"
    )
  } 
)

runServer("0.0.0.0", 9876, app, 250)
```
Let's try apache benchmark
```
$ ab http://127.0.0.1:9876/
This is ApacheBench, Version 2.3 <$Revision: 1706008 $>
Benchmarking 127.0.0.1 (be patient)...apr_pollset_poll: The timeout specified has expired (70007)
```

Let's try nagios check
```
grappa@adel /tmp/httpuv $ /usr/lib/nagios/plugins/check_http -H localhost -p 9876
CRITICAL - Socket timeout after 10 seconds
```

Let's try curl
```
$ curl -v http://127.0.0.1:9876
* Rebuilt URL to: http://127.0.0.1:9876/
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 9876 (#0)
[...]

PR #99, opened 2017-10-31
Use correct keep-alive behavior

### LINKED ISSUE #98: Parse error with persistent connections
If you run an app with `Connection = 'keep-alive'` and issue multiple requests to it, it prints a parse error.

Here's a simple test app:

```R
library(httpuv)
runServer("0.0.0.0", 5000,
  list(
    call = function(req) {
      list(
        status = 200L,
        headers = list(
          'Content-Type' = 'text/html',
          Connection = 'keep-alive'
        ),
        body = "abc"
      )
    } 
  )
)
```

To make multiple requests from a single connection, run:

```
nc localhost 5000
GET / HTTP/1.0

GET / HTTP/1.0

```

In the terminal, it'll look like this:

```
$ nc localhost 5000
GET / HTTP/1.0

HTTP/1.1 200 OK
Content-Type: text/html
Connection: keep-alive
Content-Length: 3

abcGET / HTTP/1.0
```

In the R session, it prints this  when it gets the second request:

```
ERROR: [on_request_read] parse error
```

With a git bisect, I found that this started happening in 1058ce2, which is very surprising, since it just changed the Makevars file. (This is on a Mac.)

Oddly, if you use `GET / HTTP/1.1` instead of `GET / HTTP/1.0`, it works fine in that commit, and it works fine on the current master (8d6486a).

wch:
This fixes #98. Now when either of:

* HTTP 1.0 request with no `Connection` header
* HTTP 1.1 request with a `Connection: close` header

comes in, httpuv will add a `Connection: close` **response** header and stop paying attention to incoming data.

### DIFF
--- a/src/http-parser/http_parser.c
+++ b/src/http-parser/http_parser.c
@@ -978,6 +978,11 @@ reexecute:
         if (ch == CR || ch == LF)
           break;
         parser->flags = 0;
+        /* Set headers_status in the s_start_req state, not just in
+         * http_parser_init(), because there can be multiple requests on a
+         * single connection, and we need to reset headers_status each time.
+         */
+        parser->headers_status = -1;
         parser->content_length = ULLONG_MAX;
 
         if (UNLIKELY(!IS_ALPHA(ch))) {
@@ -1869,6 +1874,10 @@ reexecute:
              * processing the headers, it's done with the http parser.
              */
             case 3:
+              /* If this is a keep-alive connection, we need to reset the state
+               * so that the parser can accept more messages.
+               */
+              UPDATE_STATE(NEW_MESSAGE());
               RETURN(p - data + 1);
 
             case 2:
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@ httpuv 1.3.5.9002
 
 * httpuv can now handle request callbacks asynchronously. ([#80](https://github.com/rstudio/httpuv/pull/80))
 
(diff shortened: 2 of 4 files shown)


src/http-parser/http_parser.c
```diff
@@ -1869,6 +1874,10 @@ size_t http_parser_execute (http_parser *parser,
              * processing the headers, it's done with the http parser.
              */
             case 3:
+              /* If this is a keep-alive connection, we need to reset the state
+               * so that the parser can accept more messages.
+               */
+              UPDATE_STATE(NEW_MESSAGE());
```
wch: This is necessary because on master, the following app would erroneously close the connection:

```R
runServer("0.0.0.0", 5000,
  list(
    onHeaders = function(req) {
      list(
        status = 200L,
        body = "abc"
      )
    }
  )
)
```

when two `GET / HTTP/1.1` requests arrived:

```
nc localhost 5000
GET / HTTP/1.1   

GET / HTTP/1.1
```

On the R side, it printed this after the second request:

```
ERROR: [on_request_read] success
```

wch:
One odd thing is that if the R application has `Connection = "keep-alive"` and then it receives a HTTP 1.0 or Connection:close request, then it responds with:

```
Connection: keep-alive
Connection: close
```

and then closes the connection. I'm not sure the user should be playing with that header, though, since we don't give them control over the close vs. keep-alive behavior.

wch:
I've pushed a commit which makes httpuv close a connection when the request is HTTP 1.0 or there is a `Connection: close`. This fixes #72.

Now this app can be used with apachebench:

```R
runServer("0.0.0.0", 5000,
  list(
    call = function(req) {
      list(
        status = 200L,
        body = "abc\n"
      )
    }
  )
)
```

An example of how to run apachebench with this app:

```
ab -n 100 -c 10 http://127.0.0.1:5000/
```

PR #106, opened 2018-01-18
Handle I/O on background thread

wch:


shalutiwari:
Ran test on docker and seeing below errors:

**ab -n 10000 -c 1000 http://127.0.0.1:5000/sync-error**
This is ApacheBench, Version 2.3 <$Revision: 1796539 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests

Test aborted after 10 failures

apr_socket_connect(): Connection reset by peer (104)
Total of 4176 requests completed


> source('test.R'); httpuv::service(Inf)
Error in execCallbacks(timeoutSecs) : 
  Evaluation error: attempt to apply non-function.
Calls: <Anonymous> -> run_now -> execCallbacks -> .Call
Execution halted

shalutiwari:
Another error: Ran on docker

> source('test.R'); httpuv::service(Inf)

Error in simpleError(msg, call) : unused arguments (function () 
{
    lapply(private$onFulfilled, function(f) {
        f(private$value, private$visible)
    })
    private$onFulfilled <- list()
}, call)
Calls: <Anonymous> ... tryCatchOne -> doTryCatch -> tryCatchList -> tryCatchOne
Execution halted
Error in (function (e)  : unused argument (<environment>)

src/callbackqueue.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include "queue.h"
+#include <boost/function.hpp>
+#include <uv.h>
+
+class CallbackQueue {
+public:
+  CallbackQueue(uv_loop_t* loop);
+  void push(boost::function<void (void)> cb);
+  void flush();
```
jcheng5: Maybe make this private

src/httpuv.cpp
```diff
@@ -12,28 +12,169 @@
...
+  // Cleanup stuff
+  uv_walk(get_io_loop(), close_handle_cb, NULL);
+  uv_run(get_io_loop(), UV_RUN_ONCE);
+  uv_loop_close(get_io_loop());
+  io_loop_initialized = false;
+}
+
+void ensure_io_thread() {
+  ASSERT_MAIN_THREAD()
+  if (io_thread_running) {
```
jcheng5: If `ensure_io_thread` is called twice in a row, you could have two bg threads created. Probably should not return from `ensure_io_thread` until the background thread is started.

src/httpuv.cpp
```diff
@@ -12,28 +12,169 @@
...
+    return;
+  }
+
+  ensure_io_loop();
+  background_queue = new CallbackQueue(get_io_loop());
```
jcheng5: It'd be good if these are called on the background thread instead, so these objects that can only be touched from the background thread are also created on the background thread.

src/httpuv.cpp
```diff
@@ -71,188 +245,130 @@ Rcpp::RObject makePipeServer(const std::string& name,
...
+  // Each call to stopServer also removes it from the pServers list.
+  while (pServers.size() > 0) {
+    stopServer(pServers[0]);
   }
-  #endif
+
+  uv_async_send(&async_stop_io_loop);
+
+  uv_thread_join(&io_thread_id);
+  io_thread_running = false;
```
jcheng5: `io_thread_running` must either only be read/written from a single thread, ever, or it needs to be guarded by a mutex

src/callbackqueue.cpp
```diff
@@ -0,0 +1,49 @@
...
+// This non-class function is a plain C wrapper for CallbackQueue::flush(), and
+// is needed as a callback to pass to uv_async_send.
+void flush_callback_queue(uv_async_t *handle) {
+  CallbackQueue* wq = reinterpret_cast<CallbackQueue*>(handle->data);
+  wq->flush();
+}
+
+
+CallbackQueue::CallbackQueue(uv_loop_t* loop) {
+  q = queue< boost::function<void (void)> >();
```
jcheng5: Remove this; q should be automatically initialized

src/callbackqueue.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+// This non-class function is a plain C wrapper for CallbackQueue::flush(), and
+// is needed as a callback to pass to uv_async_send.
+void flush_callback_queue(uv_async_t *handle) {
+  CallbackQueue* wq = reinterpret_cast<CallbackQueue*>(handle->data);
+  wq->flush();
+}
+
+
+CallbackQueue::CallbackQueue(uv_loop_t* loop) {
```
jcheng5: Add a thread assertion

src/filedatasource-unix.cpp
```diff
@@ -34,19 +37,20 @@ uint64_t FileDataSource::size() const {
...
+    throw std::runtime_error("Couldn't allocate buffer");
   }
 
   ssize_t bytesRead = read(_fd, buffer, bytesDesired);
   if (bytesRead == -1) {
-    REprintf("Error reading: %d\n", errno);
+    err_printf("Error reading: %d\n", errno);
     free(buffer);
-    throw Rcpp::exception("File read failed");
+    throw std::runtime_error("File read failed");
```
jcheng5: These could crash the process?

src/filedatasource-win.cpp
```diff
@@ -40,19 +42,21 @@ uint64_t FileDataSource::size() const {
...
   }
 
   DWORD bytesRead;
   if (!ReadFile(_hFile, buffer, bytesDesired, &bytesRead, NULL)) {
-    REprintf("Error reading: %d\n", GetLastError());
+
+    err_printf("Error reading: %d\n", GetLastError());
     free(buffer);
-    throw Rcpp::exception("File read failed");
+    throw std::runtime_error("File read failed");
```
jcheng5: Same concern as above--does this crash the thread/process?

shalutiwari:
```
Getting below error when running : ab -n 15000 -c 800 http://127.0.0.1:5000/sync-delay 

> source('test.R'); httpuv::service(Inf)
ERROR: [uv_write] bad file descriptor
```

src/http.cpp
```diff
@@ -106,13 +126,26 @@ uv_stream_t* createTcpServer(uv_loop_t* pLoop, const std::string& host,
...
+// background thread, then waits for this to finish, using a barrier.
+void createTcpServerSync(uv_loop_t* pLoop, const std::string& host,
+  int port, WebApplication* pWebApplication, CallbackQueue* background_queue,
+  uv_stream_t** pServer, uv_mutex_t* mutex, uv_cond_t* cond)
+{
+  ASSERT_BACKGROUND_THREAD()
+  *pServer = createTcpServer(pLoop, host, port, pWebApplication, background_queue);
+
+  // Tell the main thread that the server is ready
+  uv_mutex_lock(mutex);
```
jcheng5: General pattern is to put this above the work you're doing; that way, the foreground thread and background thread are not doing work concurrently (easier to reason about).

src/http.cpp
```diff
@@ -72,20 +77,35 @@ uv_stream_t* createPipeServer(uv_loop_t* pLoop, const std::string& name,
...
+  uv_cond_signal(cond);
+  uv_mutex_unlock(mutex);
+}
+
+
 uv_stream_t* createTcpServer(uv_loop_t* pLoop, const std::string& host,
-  int port, WebApplication* pWebApplication) {
+  int port, WebApplication* pWebApplication, CallbackQueue* background_queue) {
```
jcheng5: Thread assertion

src/http.h
```diff
@@ -22,21 +26,74 @@ struct Address {
...
+// As long as R has the XPtr object, the shared_ptr's target won't be deleted.
+// Also, when the XPtr gets GC'd, the shared_ptr will get deleted, and if the
+// refcount goes to 0, then the target object will be deleted (or, if it has a
+// deleter, that will be called). This means that the target object could be
+// deleted from the main thread due to a GC event in R.
+template <typename T>
+Rcpp::XPtr<boost::shared_ptr<T>,
+          Rcpp::PreserveStorage,
+          Rcpp::standard_delete_finalizer<boost::shared_ptr<T>>,
+          true> externalize_shared_ptr(boost::shared_ptr<T> obj)
```
jcheng5: Explain why the XPtr type is spelled out so explicitly--because the last argument defaults to false and we need true (finalizeOnExit)

src/http.h
```diff
@@ -22,21 +26,74 @@ struct Address {
...
+             Rcpp::PreserveStorage,
+             Rcpp::standard_delete_finalizer<boost::shared_ptr<T>>,
+             true>
+            obj_xptr(obj_copy, true);
+
+  return obj_xptr;
+}
+
+// Given an XPtr to a shared_ptr, return the shared_ptr. This does not affect
+// the shared_ptr's ref count.
```
jcheng5: The ref count does increase by one, due to the returned (copy) shared_ptr

src/http.h
```diff
@@ -22,21 +26,74 @@ struct Address {
...
+template <typename T>
+boost::shared_ptr<T> internalize_shared_ptr(
+  Rcpp::XPtr<boost::shared_ptr<T>,
+             Rcpp::PreserveStorage,
+             Rcpp::standard_delete_finalizer<boost::shared_ptr<T>>,
+             true> obj_xptr)
+{
+  boost::shared_ptr<T>* obj_copy = obj_xptr.get();
+  // Return the shared pointer's value
```
jcheng5: "Return a copy of the shared pointer"

src/http.cpp
```diff
@@ -1,35 +1,42 @@
...
 void on_request(uv_stream_t* handle, int status) {
+  ASSERT_BACKGROUND_THREAD()
   if (status) {
-    REprintf("connection error: %s\n", uv_strerror(status));
+    err_printf("connection error: %s\n", uv_strerror(status));
     return;
   }
 
-  Socket* pSocket = (Socket*)handle->data;
+  boost::shared_ptr<Socket>* pSocket = (boost::shared_ptr<Socket>*)handle->data;
```
jcheng5: I'm not sure this is right? Shouldn't it be more like:

```cpp
boost::shared_ptr<Socket> pSocket(*(boost::shared_ptr<Socket>*)handle->data);
```

I'm not sure I have the precedence right, but the point being, that `pSocket` should be a shared_ptr, not a pointer to a shared_ptr? To cause the reference count to increase while you're using it.

src/httprequest.cpp
```diff
@@ -111,8 +113,49 @@ Address HttpRequest::clientAddress() {
...
+  ASSERT_BACKGROUND_THREAD()
+
+  if (_handling_request) {
+    err_printf("Error: pipelined HTTP requests not supported.\n");
+    close();
+  }
+
+  _handling_request = true;
+  _headers.clear();
+  _response_scheduled = false;
```
jcheng5: `_bytesRead = 0;`?

src/http.cpp
```diff
@@ -38,81 +45,120 @@ void on_request(uv_stream_t* handle, int status) {
...
-  pSocket->pWebApplication = pWebApplication;
+  uv_tcp_init(pLoop, &(*pSocket)->handle.tcp);
+  (*pSocket)->handle.isTcp = true;
+  (*pSocket)->handle.stream.data = pSocket;
 
   struct sockaddr_in address = {0};
   int r = uv_ip4_addr(host.c_str(), port, &address);
   if (r) {
-    pSocket->destroy();
+    (*pSocket)->destroy();
```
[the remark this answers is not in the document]
wch: The tricky bit here is that, at present, `Socket::destroy()` schedules some callbacks, and only once the callbacks are completed does it actually delete the `Socket*`. At the moment, I'm not sure whether or not it's safe to move the content of `destroy()` into `~Socket()`. If it is safe, that simplifies the code; if not, then I think that we could add a deleter function for the shared_ptr which does basically the same thing as `destroy()` currently does.

src/httprequest.h
```diff
@@ -5,35 +5,51 @@
...
-  Rcpp::Environment _env;
-  WebSocketConnection* _pWebSocketConnection;
+  boost::shared_ptr<WebSocketConnection> _pWebSocketConnection;
+
+  // `_env` is an Environment* instead of an Environment because it must be
+  // created and deleted on the main thread. However, the creation and
+  // deletion of HttpRequest objects happens on the background thread, and so
+  // the lifetime of the Environment can't be strictly tied to the lifetime of
+  // the HttpRequest.
+  boost::shared_ptr<Rcpp::Environment> _env;
```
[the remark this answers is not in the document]
wch: In the implementation, there's a deleter function that ensures that the deletion happens on the main thread:
https://github.com/rstudio/httpuv/blob/3baaf620/src/httprequest.cpp#L146-L154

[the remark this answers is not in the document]
jcheng5: Oh nice!

src/httprequest.h
```diff
@@ -98,6 +141,27 @@ class HttpRequest : WebSocketConnectionCallbacks {
                    const char* pFooter, size_t footerSize);
   void closeWSSocket();
 
+  // Call this function from the main thread to indicate that a response has
+  // been scheduled. This is needed because sometimes by the time the main
+  // thread knows that it needs to send a response, the bg thread will have
+  // kept going and scheduled another call into the main thread to send a
+  // response.
+  void responseScheduled();
+  bool isResponseScheduled();
```
jcheng5: Don't really understand this--I'd like to talk about this and whether there are any race conditions (no mutexes)

jcheng5: Never mind, I get it now. :)

src/callback.h
```diff
@@ -0,0 +1,33 @@
...
+
+class Callback {
+public:
+  virtual ~Callback() {};
+  virtual void operator()() = 0;
+};
+
+// If the Callback class were integrated into later, this wouldn't be
+// necessary -- later could accept a void(Callback*) function.
+void invoke_callback(void* data);
```
jcheng5: Might want to add a convenience function that takes a `boost::function<void(void)>` and invokes later for you. Lots of the code in httprequest.cpp seems to follow the same pattern and would be easier to read if it was just `invoke_later(boost::bind(&method, shared_from_this(), args...))`.

src/filedatasource-unix.cpp
wch: Throwing an exception on another thread does indeed crash the process ([Example](https://gist.github.com/wch/d4f7b8654112543217184d2b3202a88f)). What do you think it should do instead?

wch:
Things to do before merging:
- [ ] Run with shinyloadtest
- [ ] automated load testing in Travis

PR #113, opened 2018-03-13
Avoid NULL WebSocketConnection

### LINKED ISSUE #112: Reloading a Shiny application can cause a segfault
Using the development version of httpuv and shiny, run an app and then reload many times quickly. For debugging info, launch R with gdb or lldb:

```R
R -d lldb
run
library(shiny)
runExample('01_hello', display.mode="normal", launch.browser=T)
```

It stops here:

```
* thread #2, stop reason = EXC_BAD_ACCESS (code=1, address=0x20)
    frame #0: 0x0000000108698fe8 httpuv.so`WebSocketConnection::read(this=0x0000000000000000, data=0x0000000000000000, len=0) at websockets.cpp:294
   291  
   292  void WebSocketConnection::read(const char* data, size_t len) {
   293    ASSERT_BACKGROUND_THREAD()
-> 294    assert(_pParser);
   295    _pParser->read(data, len);
   296  }
   297  
Target 0: (R) stopped.
(lldb) 
```

The cause of the problem is that the HttpRequest object has been closed, but there is a scheduled `WebSocketConnection::read` that executes afterward.

wch:
This fixes #112. It checks that `_pWebSocketConnection` is not null before scheduling operations with it. It also ensures that the `WebSocketConnection` objects are deleted on the background thread.

There is also now a `WebSocketConnection::markClosed()` method. This can be called to indicate that the connection has been closed without going through the normal sending and receiving of a close message.

There's one more change that's tangential: it converts `vector<char>*` to `shared_ptr`s, which means that the deletion doesn't have to be manually scheduled.

### DIFF
--- a/src/httprequest.cpp
+++ b/src/httprequest.cpp
@@ -583,6 +583,11 @@ void HttpRequest::closeWSSocket() {
 void HttpRequest::_on_closed(uv_handle_t* handle) {
   ASSERT_BACKGROUND_THREAD()
   trace("HttpRequest::_on_closed");
+  // Tell the WebSocketConnection that the connection is closed, before
+  // resetting the shared_ptr. This is useful because there may be some
+  // callbacks that will execute later, and we want to make sure the WSC
+  // doesn't try to do anything with them.
+  _pWebSocketConnection->markClosed();
   _pWebSocketConnection.reset();
 }
 
--- a/src/httprequest.cpp
+++ b/src/httprequest.cpp
@@ -1,5 +1,6 @@
 #include <boost/bind.hpp>
 #include <boost/shared_ptr.hpp>
+#include <boost/make_shared.hpp>
 #include <later_api.h>
 #include "httprequest.h"
 #include "callback.h"
@@ -361,7 +362,7 @@ int HttpRequest::_on_body(http_parser* pParser, const char* pAt, size_t length) 
 
   // Copy pAt because the source data is deleted right after calling this
   // function.
-  std::vector<char>* buf = new std::vector<char>(pAt, pAt + length);
+  boost::shared_ptr<std::vector<char>> buf = boost::make_shared<std::vector<char>>(pAt, pAt + length);
 
   boost::function<void(boost::shared_ptr<HttpResponse>)> schedule_bg_callback(
     boost::bind(&HttpRequest::_schedule_on_body_error, shared_from_this(), _1)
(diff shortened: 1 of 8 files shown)


src/websockets.cpp
```diff
@@ -343,12 +348,17 @@ void WebSocketConnection::onFrameComplete() {
...
+          _connState = WS_CLOSE_RECEIVED;
+        } else if (_connState == WS_CLOSE_SENT) {
+          _connState = WS_CLOSED;
+        }
+        
         // If we haven't sent a Close frame before, send one now, echoing
         // the callback
-        if (!(_connState & WS_CLOSE_SENT)) {
-          _connState |= WS_CLOSE_SENT;
+        if (_connState == WS_CLOSE_RECEIVED) {
```
wch: The intent would be clearer if we test against `WS_CLOSE_SENT` and `WS_CLOSED`.

src/httprequest.cpp
```diff
@@ -653,7 +653,13 @@ void HttpRequest::_call_r_on_ws_open() {
   boost::function<void (void)> error_callback(
     boost::bind(&HttpRequest::schedule_close, shared_from_this())
   );
-  this->_pWebApplication->onWSOpen(shared_from_this(), error_callback);
+  // Passing p_wsc separately is necessary, even though
+  // shared_from_this()->websocket() is nominally the same thing. This is
+  // because there's a potential race where shared_from_this()->websocket() is
+  // deleted on another thread, and then the onWSOpen method tries to access
+  // the null pointer. Passing p_wsc guarantees that it won't get deleted
```
wch: "deleted" isn't right -- use reset() instead.

src/webapplication.cpp
```diff
@@ -307,13 +307,14 @@ void RWebApplication::getResponse(boost::shared_ptr<HttpRequest> pRequest,
   UNPROTECT(1);
 }
 
-void RWebApplication::onWSOpen(boost::shared_ptr<HttpRequest> pRequest,
+void RWebApplication::onWSOpen(boost::shared_ptr<WebSocketConnection> pConn,
+                               boost::shared_ptr<HttpRequest> pRequest,
                                boost::function<void(void)> error_callback) {
   ASSERT_MAIN_THREAD()
```
wch: Don't take `pConn` as an argument. Instead, acquire the `->websocket()` and check that it's not null here.

src/httprequest.cpp
```diff
@@ -653,7 +653,13 @@ void HttpRequest::_call_r_on_ws_open() {
...
+  // before it's used.
+  this->_pWebApplication->onWSOpen(p_wsc, shared_from_this(), error_callback);
 
   // _requestBuffer is likely empty at this point, but copy its contents and
   // _pass along just in case.
```
wch: Should acquire `_pWebSocketConnection` here and check for non-null. (don't acquire above.)

jcheng5:
I've been putting the branch through its paces myself, no crashes but I did get one `ERROR: [on_request_read] connection reset by peer`, didn't we remove those already?

wch:
We never explicitly got rid of that error -- it's that the code was restructured in a way that it didn't happen (often). I do recall discussing the possibility of suppressing that message though, since it can be needlessly alarming. I can do that in this branch.

wch:
Good idea - I've implemented that change.

jcheng5:
LGTM!

PR #135, opened 2018-05-08
Add lptread parameter to the build for ppc64le

barthy1:
Without lpthread parameter specification, build of httpuv is failed on Power architecute (ppc64le)

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -19,6 +19,10 @@ endif
 ifeq ($(UNAME), OpenBSD)
 PKG_LIBS += -lkvm
 endif
+UNAME_ARCH=$(shell uname -m)
+ifeq ($(UNAME_ARCH), ppc64le)
+PKG_LIBS += -lpthread
+endif
 
 PKG_CPPFLAGS = $(C_VISIBILITY)
 
--- a/src/Makevars
+++ b/src/Makevars
@@ -9,7 +9,7 @@ ifeq ($(UNAME), Darwin)
 FRAMEWORK = -framework CoreServices
 endif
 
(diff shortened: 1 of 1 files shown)


wch:
This flag should probably be added for all platforms, according to https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-pthreads

One thing that was not clear to me is when to use `-pthread` vs `-lpthread`, but I found this SO topic, which was helpful: https://stackoverflow.com/questions/41267121/pthread-lpthread-and-minimal-dynamic-linktime-dependencies

This is also related to r-lib/later#11.

barthy1:
@wch thank you for the links. How would you recommend to update the pr? I can update the code to add -lpthread for all platforms, but I cannot test it on all these platfrom. Also if it works without extra parameter, why do we need to add it?

wch:
I'm not sure exactly what to do here. I believe that for gcc and clang, it should be `-pthread`, and for other compilers, it should be `-lpthread`. The documentation doesn't say how to tell which compiler is being used.

I think that the simplest thing to do is to just copy what we do in the later package:
https://github.com/r-lib/later/blob/207d4e01/src/Makevars#L1-L2

@barthy1 If you set those flags for httpuv, does it work for you?

barthy1:
@wch tested with `-pthread` for PKG_CPPFLAGS and PKG_LIBS, it works ok. So updated the PR accordingly, as we dicussed put the parameter for all platforms.

wch:
@barthy1 Can you sign and send a contributor agreement? The information is at https://github.com/rstudio/httpuv/blob/master/CONTRIBUTING.md.

Or, if you prefer, we can implement the `-pthread` fix ourselves.

barthy1:
@wch I'm ok if you implement that. Thank you

wch:
I've implemented this in 2c5634a.

Issue #116, opened 2018-03-30
Calling stopAllServers() can result in crash


wch:
```
R -d lldb
run --quiet
library(httpuv)
startServer("0.0.0.0", 5001, list(call = function(req) {}))
startServer("0.0.0.0", 5001, list(call = function(req) {}))
stopAllServers()
```

Result:

```
$ R -d lldb
*** Further command line arguments ('--no-save --no-restore-data --quiet ') disregarded
*** (maybe use 'run --no-save --no-restore-data --quiet ' from *inside* lldb)

(lldb) target create "/Library/Frameworks/R.framework/Resources/bin/exec/R"
Current executable set to '/Library/Frameworks/R.framework/Resources/bin/exec/R' (x86_64).
(lldb) run --quiet
Process 40059 launched: '/Library/Frameworks/R.framework/Resources/bin/exec/R' (x86_64)
> library(httpuv)
> startServer("0.0.0.0", 5001, list(call = function(req) {}))
[1] "4330444536"
> startServer("0.0.0.0", 5001, list(call = function(req) {}))
Error in startServer("0.0.0.0", 5001, list(call = function(req) { : 
  Failed to create server
> stopAllServers()
Assertion failed: (0), function uv_close, file src/unix/core.c, line 182.
Process 40059 stopped
* thread #2, stop reason = signal SIGABRT
    frame #0: 0x00007fff566e6e3e libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff566e6e3e <+10>: jae    0x7fff566e6e48            ; <+20>
    0x7fff566e6e40 <+12>: movq   %rax, %rdi
    0x7fff566e6e43 <+15>: jmp    0x7fff566de0b8            ; cerror_nocancel
    0x7fff566e6e48 <+20>: retq   
Target 0: (R) stopped.
(lldb) bt
* thread #2, stop reason = signal SIGABRT
  * frame #0: 0x00007fff566e6e3e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff56825150 libsystem_pthread.dylib`pthread_kill + 333
    frame #2: 0x00007fff56643312 libsystem_c.dylib`abort + 127
    frame #3: 0x00007fff5660b368 libsystem_c.dylib`__assert_rtn + 320
    frame #4: 0x0000000107e7c492 httpuv.so`uv_close + 434
[...]

Issue #134, opened 2018-05-07
Sanitizer complains when starting app after a failed app start


wch:
This happens with the RDsan build of R from [wch1/r-debug](https://hub.docker.com/r/wch1/r-debug/). It doesn't happen for me on my Mac. This might be related to #133.

Minimal example:

```R
docker run --rm -ti wch1/r-debug

RDsan
install.packages('httpuv') # Currently 1.4.2
library(httpuv)

startServer("0.0.0.0", 8000, list())
startServer("0.0.0.0", 8000, list())
startServer("0.0.0.0", 8001, list()) # Crashes here. Will also crash with 8000.
```

It happens when you start an app, try to start another app on the same port (which fails), and then try to start another app.

Output:

```
> library(httpuv)
> startServer("0.0.0.0", 8000, list())
[1] "106858786783336"
> startServer("0.0.0.0", 8000, list())
Error in startServer("0.0.0.0", 8000, list()) : Failed to create server
> startServer("0.0.0.0", 8000, list())
=================================================================
==32185==ERROR: AddressSanitizer: heap-use-after-free on address 0x613000070248 at pc 0x7f35781d4759 bp 0x7f3577d10ea0 sp 0x7f3577d10e90
WRITE of size 8 at 0x613000070248 thread T1
    #0 0x7f35781d4758 in uv__stream_init src/unix/stream.c:76
    #1 0x7f35781ddcc1 in uv_tcp_init_ex src/unix/tcp.c:127
    #2 0x7f35781dde1e in uv_tcp_init src/unix/tcp.c:146
    #3 0x7f35780cf8ad in createTcpServer(uv_loop_s*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, boost::shared_ptr<WebApplication>, CallbackQueue*) /tmp/RtmpIOOdMu/R.INSTALL47bf24cd26c2/httpuv/src/http.cpp:120
    #4 0x7f35780d0321 in createTcpServerSync(uv_loop_s*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, boost::shared_ptr<WebApplication>, CallbackQueue*, uv_stream_s**, Barrier*) /tmp/RtmpIOOdMu/R.INSTALL47bf24cd26c2/httpuv/src/http.cpp:172
[...]

PR #138, opened 2018-05-09
wip for #134

jcheng5:
AddressSanitizer was right, this appears to be a legitimate use-after-free. When `uv_tcp_init` is called, it adds a reference to the handle to `uv_loop_t`. https://github.com/rstudio/httpuv/blob/248723576422281b9a039b336f2df3cff6967d20/src/libuv/src/uv-common.h#L216

This reference doesn't get removed until after `uv_close` asynchronously completes. https://github.com/rstudio/httpuv/blob/5d1b7efc01dabfb373ee5e71830e59a2bde30236/src/libuv/src/unix/core.c#L283

We were not calling `uv_close` in case of error. The subsequent call to `uv_tcp_init` causes the next tcp handle to be added to `uv_loop_t`, and traversing the list of references causes the sanitizer alarms to be tripped.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@ httpuv 1.4.2.9000
 
 * Fixed [#133](https://github.com/rstudio/httpuv/issues/133): Assertion failures when running on Fedora 28. ([#136](https://github.com/rstudio/httpuv/pulls/136))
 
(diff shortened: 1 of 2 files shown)


wch:
I haven't tested yet, but it seems likely that this also fixes #116.

This looks good to me, though we should discuss in person. I want to be sure that the lifetimes of the Socket objects are managed correctly.

For reference, the lines of code that you replaced were added in 44573b5.

PR #142, opened 2018-05-24
Export ipFamily() function

wch:
This is so that we can fix rstudio/shiny#2074.


src/httpuv.cpp
```diff
@@ -599,6 +599,36 @@ std::vector<std::string> decodeURIComponent(std::vector<std::string> value) {
...
+//' @export
+// [[Rcpp::export]]
+int ipFamily(const std::string& ip) {
+  int family = ip_family(ip);
+  if (family == AF_INET6)
+    return 6;
+  else if (family == AF_INET)
+    return 4;
+  else
+    return -1;
```
wch: Maybe this should return `NA` instead of `-1`? If we do that, we'll have to also make changes to the PR on Shiny, because it uses `if (ipFamily(x) == 6)`.

jcheng5:
I hate to do this to you. But you've opened the pandora's box of IPv6.

http://veithen.github.io/2013/12/30/how-to-correctly-parse-ipv6-addresses.html

FWIW, zone IDs aren't supported by browsers (yet) so we can proceed like this if you want. Another way to go would be to leave httpuv unchanged and add a regex to Shiny. I think all that's really required is to look for `:`...?

wch:
After a lot of reading and learning about zone IDs, I decided to look at the libuv sources for `uv_inet_pton`, which is what the `ip_family` function uses. It turns out it already knows about zone IDs.

https://github.com/rstudio/httpuv/blob/2c5634a/src/libuv/src/inet.c#L168

```R
# Zone ID valid for IPv6
ipFamily("fe80::91:58b5:470c:7c54%en0")
# [1] 6
ipFamily("::1%lo0")
# [1] 6
ipFamily("::1%1")
# [1] 6
ipFamily("::31%")
# [1] 6

# Zone ID not valid for IPv4
ipFamily("1.1.1.1")
# [1] 4
ipFamily("1.1.1.1%2")
# [1] -1
```

PR #143, opened 2018-05-24
Add HEADERS variable to the Rook request environment

### LINKED ISSUE #141: Get raw headers from request
Currently, callback functions registered via `call` or `onHeaders` can only receive the "Rook-style" request headers ("HTTP_ Variables" in the [Rook README](https://github.com/jeffreyhorner/Rook)).

It would be useful if there was a way to receive the raw headers, before they were munged to become variable names, perhaps as a list of pairs of strings. That way, headers could be inspected without accounting for Rook's renaming of them.

Interested parties so far include @wch and @jeroen.

This capability raises various questions.

## Questions

1. What value should be used to represent raw headers?
    1. Shouldn't be an `env` because the same header might appear multiple times. This is valid in HTTP under [certain circumstances](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)
1. How should a handler be registered so that httpuv knows to pass it raw headers? And, how should that value be passed? Possibilities include:
    1. New `callRaw` and `onHeadersRaw` `app` members that pass the handler a second argument, the raw headers value. Using in conjunction with `call`/`onHeaders` respectively is an error.
    2. Always add the raw headers as an attribute of the Rook `req` object passed to `call`/`onHeaders`
        1. Problematic because attributes can contribute to object equality. Might not be backward-compatible?
        1. Introduces overhead we don't want to pay all the time?
    3. Add an `addRawRequest = TRUE` (default `FALSE`) member to `app`. This would be the only member of `app` so far that's not a function. Causes raw header value to be added to all `req`s as an attribute.
    4. Like above, but add the `addRawRequest = TRUE` optional argument to `startServer`/`startPipeServer` instead of as `app` member.
    5. Add `req$headers`
        1. Should this be an addition to Rook first, then httpuv?

alandipert:
Fixes #141 

TODO:
- [x] NEWS
- [x] Reference documentation


wch:
Some notes:

* The `req` object now includes an item `req$HEADERS`, which is a named character vector.
* The names are all lowercase, since the HTTP spec says that capitalization does not matter, and making it all lowercase makes it easier to consistently index into.
* The order of headers is not preserved, also in keeping with the HTTP spec.
* If multiple headers have the same name, as in:
    ```
    foo: 123
    FOo: 456
   ```
    then they will be collapsed into a single value (`req$HEADERS[["foo"]] == "123,456"`), separated by a comma, and preserving the order. This is also what the HTTP spec says.

jeroen:
Thanks this is great! Thank you!

jeroen:
Make sure you test this on CentOS 6 though, some old compilers have trouble with iterators.

alandipert:
@jeroen thank you for the feedback, I'm glad you like it!

Since making the PR we've become somewhat ambivalent about adding this now, since the info was already available, albeit in an inconvenient form. We (@jcheng5 and @wch and I) discussed passing for now, and instead holding out for a possible overhaul of the httpuv API that could include it.

What are you using it for? Would this be really helpful to you now, or would you be willing to wait?

Thanks in advance for sharing your perspective.

alandipert:
Thanks for your input @jeroen, scratch what I said, we're going to move forward with this 😄  I'll do the CentOS 6 testing in the very near future.

jeroen:
Just checked in my centos 6 vm and all seems good !

alandipert:
@wch when you have a chance can you please look at the NEWS and documentation additions and let me know if they're good?

PR #154, opened 2018-06-08
Require later >=0.7.3

jcheng5:
Don't merge until later 0.7.3 goes to CRAN


wch:
Later 0.7.3 is on CRAN, so there's no need to have it listed in Remotes.

PR #160, opened 2018-06-23
Fix Windows build for gcc 8.1.0

jeroen:
The new Windows toolchain requires that `libuv` is compiled with ` -D_WIN32_WINNT=0x0600`. I also cleaned up some other flags while I was at it.

### DIFF
--- a/src/Makevars.win
+++ b/src/Makevars.win
@@ -4,11 +4,10 @@ CXX_STD=CXX11
 PKG_LIBS = ./libuv/libuv.a ./http-parser/http_parser.o ./sha1/sha1.o ./base64/base64.o \
 	-lpthread -lws2_32 -lkernel32 -lpsapi -liphlpapi -lshell32 -luserenv
 
-PKG_CPPFLAGS += -D_WIN32_WINNT=0x0600 -DSTRICT_R_HEADERS
-CFLAGS += -D_WIN32_WINNT=0x0600 -DSTRICT_R_HEADERS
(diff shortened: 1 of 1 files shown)


wch:
Thanks! Is gcc 8.1.0 going into Rtools soon? Really what I'm wondering is if we'll need to do a release soon.

jeroen:
I'll let you know. I'm trying to push but things always take longer than expected.

jeroen:
FWIW an interesting technical detail is that this broke due to the toolchain update of `make.exe` rather than the new version of `gcc`. It seems the new make has slightly different behavior wrt inheritance of make variables within submakes.

wch:
Do you have any references to the change in make's behavior? I wonder if this is something we need to watch out for Unix builds as well.

jeroen:
Not really. The problem is that the make from the old rtools was 15 years old so a thing or two might have changed. I don't think you have to worry about it on other systems.

I suspect (but haven't verified this yet) that the change is related specifically to the treatment of the `CPPFLAGS` variable as part of the [implicit make rules](https://www.gnu.org/software/make/manual/html_node/Implicit-Rules.html).

jeroen:
Actually if you could push out a release one of these days that would be really helpful. The toolchain is ready and I want to ask package authors to start testing their packages. But many packages won't be able to install if httpuv can't be installed.

Also CRAN will not deploy this toolchain as long as there are too many packages failing...

Sorry for the hassle.

wch:
We just submitted a new version to CRAN. Hopefully it'll go through!

jeroen:
Awesome thanks for the quick turnaround!

PR #153, opened 2018-06-08
Be lenient about Connection header contents when Upgrade header is present

jcheng5:
This is a workaround for https://github.com/rstudio/rstudio/issues/2940. It
stops an erroneously sent "Connection: close" header from breaking an
otherwise legit websocket request.

PR #162, opened 2018-07-03
Make sure connection is not upgraded too early

### LINKED ISSUE #161: Shiny apps sometimes crash on rstudio.cloud with httpuv 1.4.4.2
On rstudio.cloud, running `shiny::runExample('01_hello', display.mode="normal")` results in a crash about 25-50% of the time. (This is with Chrome on Mac.)

I bisected and the bad commit seems to be b8969775600ddd8225764840c05e27274c87ca72, which was in PR #153. That fixed one issue but seems to have introduced a different one.

wch:
This fixes #161. This makes sure that a HTTP connection is not upgraded too early, before all the request headers have been processed.

In the old http-parser code, `parser->upgrade` was set in the `s_headers_almost_done` stage (https://github.com/rstudio/httpuv/blob/v1.3.6.2/src/http-parser/http_parser.c#L1567-L1568). This happens when all the request headers have been processed.

Also, in the current version of http-parser, `parser->upgrade` is set in `s_headers_almost_done` (https://github.com/rstudio/httpuv/blob/c3978bad/src/http-parser/http_parser.c#L1822-L1825) or in `s_on_headers_completed` (https://github.com/rstudio/httpuv/blob/c3978bad/src/http-parser/http_parser.c#L1884).

In #153, we added `isUpgrade()`. The problem is that it could return `true` before all the headers have been processed. This PR makes sure that `isUpgrade()` will return `true` only after all the headers have been processed.

One thing that I'm not 100% sure about: In the http-parser code, `parser->upgrade` could be set in a slightly earlier stage (`s_headers_almost_done`) than what I've done here (`HttpRequest:: _on_headers_complete`), and I wonder if that could result in different behavior.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: httpuv
 Type: Package
 Title: HTTP and WebSocket Server Library
-Version: 1.4.4.2
+Version: 1.4.4.9001
 Author: Joe Cheng, Hector Corrada Bravo [ctb], Jeroen Ooms [ctb],
     Winston Chang [ctb]
 Copyright: RStudio, Inc.; Joyent, Inc.; Nginx Inc.; Igor Sysoev; Niels Provos;
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+httpuv 1.4.4.9001
+==============
+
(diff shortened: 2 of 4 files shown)


vsrdharca:
@wch Hello Winston -

This is an old issue but even though I have most recent shiny and httpuv packages I am encountering a websocket error.
running:
shiny - 1.7.1
httpuv - 1.6.3.9000
rstudio cloud - RStudio Server Version 1.3.959

I run shiny::runExample("01_hello") and get the grey screen with the websocket connection error as in the attached image. on an older version of rstudio server 1.2.5042 I don't have get this error. I would appreciate any thoughts you might have about resolving this.

Much thanks!

image

PR #177, opened 2018-11-28
Add support for serving static files on background thread

wch:
This PR adds support for serving files on the background thread that handles I/O.

This minimal example will serve static files out of a subdir of the current directory named `content`. If there's an index.html file, it will be served if the request is for `/` (the same is true for subdirectories). If the file isn't found, a minimal 404 page will be served, without hitting the main R thread.

```R
s <- startServer("127.0.0.1", 5000,
  list(
    staticPaths = list("/" = "content")
  )
)
```

In the next example, it will try to serve requests that have `/assets` in the URL path from the local `assets` subdirectory. If the file isn't found, the request will "fall through" -- that is, it will be forwarded to the R `call` function, like requests normally are with httpuv.

```R
s <- startServer("127.0.0.1", 5000,
  list(
    call = function(req) {
      return(list(
        status = 404,
        body = paste0("404 file not found: ", req$PATH_INFO)
      ))
    },
    staticPaths = list("/assets" = "assets"),
    staticPathOptions = staticPathOptions(
      indexhtml = TRUE,
      fallthrough = TRUE
    )
  )
)
```


R/server.R
```diff
@@ -0,0 +1,211 @@
...
+        private$appWrapper$.staticPaths,
+        private$appWrapper$.staticPathOptions
+      )
+
+      if (is.null(private$handle)) {
+        stop("Failed to create server")
+      }
+
+      private$running <- TRUE
+      registerServer(self)
```
wch: Registration should happen in the `Server` class, to mirror deregistration. Same for `PipeServer`.

R/static_paths.R
```diff
@@ -0,0 +1,283 @@
...
+    "  Local path:        ", x$path, "\n",
+    format_opts(x$options)
+  )
+}
+
+#' Create options for static paths
+#'
+#'
+#' @param indexhtml If an index.html file is present, should it be served up
+#'   when the client requests \code{/} or any subdirectory?
```
wch: `/` isn't correct. Should say something like "the directory".

src/staticpath.cpp
```diff
@@ -0,0 +1,383 @@
...
+      return std::pair<StaticPath, std::string>(*sp, post_slash);
+    }
+
+    if (found_idx == 0) {
+      // We get here after checking the leading '/'.
+      return boost::none;
+    }
+
+    // Split the string on '/'
+    found_idx = path.find_last_of('/', found_idx - 1);
```
jcheng5: The first time this is called, `found_idx` is `std::string::npos`; is it OK to subtract 1 from that?

src/webapplication.cpp
```diff
@@ -173,8 +189,11 @@ boost::shared_ptr<HttpResponse> listToResponse(
   // - body: Character vector (which is charToRaw-ed) or raw vector, or NULL
   if (std::find(names.begin(), names.end(), "bodyFile") != names.end()) {
     FileDataSource* pFDS = new FileDataSource();
-    pFDS->initialize(Rcpp::as<std::string>(response["bodyFile"]),
-        Rcpp::as<bool>(response["bodyFileOwned"]));
+    int ret = pFDS->initialize(Rcpp::as<std::string>(response["bodyFile"]),
+                               Rcpp::as<bool>(response["bodyFileOwned"]));
+    if (ret != 0) {
+      REprintf(pFDS->lastErrorMessage().c_str());
```
jcheng5: Is it correct that there's no return or anything here? We just plow forward with this pFDS?

src/webapplication.cpp
```diff
@@ -364,3 +405,171 @@ void RWebApplication::onWSClose(boost::shared_ptr<WebSocketConnection> pConn) {
...
+  ASSERT_BACKGROUND_THREAD()
+
+  // If it has a Connection: Upgrade header, don't try to serve a static file.
+  // Just fall through, even if the path is one that is in the
+  // StaticPathManager.
+  if (pRequest->hasHeader("Connection", "Upgrade", true)) {
+    return nullptr;
+  }
+
+  std::string url = doDecodeURI(pRequest->url(), true);
```
jcheng5: I think you might need to do the URI decoding only after splitting on the query string. You don't want to split on an escaped `?`, only a raw one.

src/webapplication.cpp
```diff
@@ -364,3 +405,171 @@ void RWebApplication::onWSClose(boost::shared_ptr<WebSocketConnection> pConn) {
...
+
+  // Self-frees when response is written
+  FileDataSource* pDataSource = new FileDataSource();
+  int ret = pDataSource->initialize(local_path, false);
+
+  if (ret != 0) {
+    // Couldn't read the file
+    delete pDataSource;
+
+    if (sp.options.fallthrough.get()) {
```
jcheng5: Not sure we should fall through here. If a file/directory is missing it's one thing, but if it's present but we fail when reading it, that feels like a true error (and maybe 500 instead of 404).

src/webapplication.cpp
```diff
@@ -364,3 +405,171 @@ void RWebApplication::onWSClose(boost::shared_ptr<WebSocketConnection> pConn) {
...
+    }
+  }
+
+  // Set the Content-Length here so that both GET and HEAD requests will get
+  // it. If we didn't set it here, the response for the GET would
+  // automatically set the Content-Length (by using the FileDataSource), but
+  // the response for the HEAD would not.
+  respHeaders.push_back(std::make_pair("Content-Length", toString(file_size)));
+  respHeaders.push_back(std::make_pair("Content-Type", content_type));
+  respHeaders.push_back(std::make_pair("Date", http_date_string(time(NULL))));
```
jcheng5: Should we be doing this in some more-global code path? (I'm assuming it doesn't happen for HTTP requests that are serviced by R?)

src/webapplication.cpp
wch: Implemented in f3e487d and 4f499a8.

src/webapplication.cpp
wch: Yeah, the existing behavior was to do that. IIRC, it resulted in an error being thrown in `getData()`, and this message: `Error reading: <filename>`. If we fix this, I suggest doing it in a separate PR.

src/webapplication.cpp
wch: Good idea. I moved it into the constructor for `HttpResponse`.

PR #178, opened 2018-12-04
Convert from ref classes to R6

wch:
This PR makes httpuv stop using ref classes and switches it to R6 instead.


jcheng5:
Any idea why we're failing Travis?

wch:
I think it was a problem trying to upgrade callr. Clearing the Travis cache seems to have fixed it.

PR #180, opened 2018-12-14
Implement Last-Modified response header and use If-Modified-Since request header for static files

wch:
When serving static files, this adds a `Last-Modified` header to responses. It also adds the ability to use the `If-Modified-Since` header from HTTP requests, and sends back a 304 response if appropriate.

Note that in my testing, it's hard to get a browser to actually send the `If-Modified-Since` header in real-world usage. In Chrome, if you uncheck "Disable Cache" in the network panel (of the dev tools), then reload the page, it will simply retrieve the files from the browser's memory cache. I think if you wait for a long enough period of time (at least several minutes), it will make a request with `If-Modified-Since` and then receive a 304 response.

For better caching, I think there needs to be `Cache-Control` and `ETag` headers, as described here: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#validating_cached_responses_with_etags

This PR also changes all instances of `DataSource*` to a `shared_ptr<DataSource>` (and same for `FileDataSource*` and `InMemoryDataSource*`) because there were some places where managing the lifetime was getting pretty complicated. Switching to a `shared_ptr` simplifies it greatly.


src/utils.h
```diff
@@ -246,4 +247,56 @@ inline std::string http_date_string(const time_t& t) {
   return std::string(res);
 }
 
+
+// Given a date string of format "Wed, 21 Oct 2015 07:28:00 GMT", return a
+// time_t representing that time. If the date is malformed, then return 0.
+inline time_t parse_http_date_string(const std::string& date) {
```
jcheng5: Maybe test that the length of the date string is what you expect? So that the `.substr` calls are guaranteed not to throw

src/webapplication.cpp
```diff
@@ -506,14 +511,10 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
     }
   }
 
-  // Self-frees when response is written
-  FileDataSource* pDataSource = new FileDataSource();
+  boost::shared_ptr<FileDataSource> pDataSource = boost::make_shared<FileDataSource>();
```
jcheng5: Wow, `shared_ptr` here is a big improvement

src/utils.h
jcheng5: Let's assert background thread; if we end up calling this from multiple threads we should not use the static for std::locale.

wch: Good call.

PR #185, opened 2019-01-24
decodeURI and decodeURIComponent now return UTF-8 strings

### LINKED ISSUE #122: decodeURI and decodeURIComponent should return UTF-8 encoded string
For example, when running [these tests](https://github.com/rstudio/httpuv/blob/3861ebc46df9df3d82c2562d4d253123f57502d6/tests/testthat/test-utils.R) with `decodeURI` and `decodeURIComponent` on Windows, it returns a native-encoded string, but it should be UTF-8. This does match the behavior of R's `URLdecode` function, but I don't think it's right.

As for `encodeURI` and `encodeURIComponent`, I'm not sure at the moment what they should do if the input is not UTF-8 encoded.

wch:
This fixes #122.

The `decodeURI()` and `decodeURIComponent()` functions previously returned strings that had the native encoding; they now return UTF-8 strings.


wch:
I looked, and there doesn't seem to be.

hadley:
FWIW I think returning `std::vector<std::string>` from Rcpp function is bad practice for exactly this reason. You're better off returning a `CharacterVector` and explicitly using `Rf_mkCharCE(x, CE_UTF8)` to construct the individual components.

PR #192, opened 2019-02-05
URI encoding/decoding fixes

wch:
This is a follow-up to #185, and it does what @hadley suggested: mark the encoding of characters in C++ instead of R.

It also ensures that vector inputs and `NA`s are handled properly.


src/httpuv.cpp
```diff
@@ -501,60 +501,66 @@ std::string doEncodeURI(std::string value, bool encodeReserved) {
...
-    it++) {
+Rcpp::CharacterVector encodeURI(Rcpp::CharacterVector value) {
+  Rcpp::CharacterVector out(value.size());
 
-    *it = doEncodeURI(*it, false);
+  for (int i = 0; i < value.size(); i++) {
+    if (value[i] == NA_STRING) {
+      out[i] = NA_STRING;
+    } else {
+      const char* s = doEncodeURI(Rcpp::as<std::string>(value[i]), false).c_str();
```
hadley: Don't you have an encoding problem here too? What does `doEncodeURI` assume about the input? You might want a helper like this:

```cpp
inline const char* string_utf8(SEXP x, int i) {
  return Rf_translateCharUTF8(STRING_ELT(x, i));
}
```

src/httpuv.cpp
```diff
@@ -501,60 +501,66 @@ std::string doEncodeURI(std::string value, bool encodeReserved) {
...
-  for (std::vector<std::string>::iterator it = value.begin();
-    it != value.end();
-    it++) {
+Rcpp::CharacterVector encodeURI(Rcpp::CharacterVector value) {
+  Rcpp::CharacterVector out(value.size());
 
-    *it = doEncodeURI(*it, false);
+  for (int i = 0; i < value.size(); i++) {
+    if (value[i] == NA_STRING) {
+      out[i] = NA_STRING;
```
hadley: You might want to see if the CharacterVector is already filled with `NA`s.

wch: It looks like `CharacterVector(n)` will return a vector filled with `""`, but `CharacterVector(n, NA_STRING)` returns a string filled with `NA`s.

src/httpuv.cpp
wch: Yes, that's true. Our current documentation for `encodeURI` says the following, but it's not generally useful behavior:

>  If conformant non-ASCII behavior is important, ensure that your input vector is UTF-8 encoded before calling encodeURI or encodeURIComponent.

This differs from the behavior of `base::URLencode`. I agree that we should change it.

```R
utf8_str <- "\ue5"   # "å", in UTF-8
latin1_str <- iconv(utf8_str, "UTF-8", "latin1")

utf8_str
#> [1] "å"
latin1_str
#> [1] "å"

# Look at raw bytes
charToRaw(utf8_str)
#> [1] c3 a5
charToRaw(latin1_str)
#> [1] e5

# base::URLencode
URLencode(utf8_str)
#> [1] "%C3%A5"
URLencode(latin1_str)
#> [1] "%C3%A5"

# httpuv::encodeURI
httpuv::encodeURI(utf8_str)
#> [1] "%C3%A5"
httpuv::encodeURI(latin1_str)
#> [1] "%E5"
```

src/httpuv.cpp
```diff
@@ -501,60 +501,62 @@ std::string doEncodeURI(std::string value, bool encodeReserved) {
...
-
-    *it = doEncodeURI(*it, false);
+Rcpp::CharacterVector encodeURI(Rcpp::CharacterVector value) {
+  Rcpp::CharacterVector out(value.size(), NA_STRING);
+
+  for (int i = 0; i < value.size(); i++) {
+    if (value[i] != NA_STRING) {
+      const char* s = Rf_translateCharUTF8(value[i]);
+      s = doEncodeURI(s, false).c_str();
+      out[i] = Rf_mkCharCE(s, CE_UTF8);
```
wch: I don't think this needs a `PROTECT`, but if someone else could confirm, I'd appreciate it.

hadley: It should not because you're immediately assigning into an object that Rcpp should be PROTECTing.

PR #193, opened 2019-02-06
Validate header in constant time

### LINKED ISSUE #191: Implement constant-time header equality check
https://github.com/rstudio/httpuv/blob/1ca8f15c5e9f5a40d88c70a9d713c9dc34b6f278/src/staticpath.cpp#L131

It'd be good if this comparison was constant time, to protect shiny-shared-secret from timing attacks.

wch:
Closes #191.

Note that the the header validation code has tests at https://github.com/rstudio/httpuv/blob/1ca8f15c/tests/testthat/test-static-paths.R#L231-L314, and they still pass.

### DIFF
--- a/src/staticpath.cpp
+++ b/src/staticpath.cpp
@@ -128,7 +128,7 @@ bool StaticPathOptions::validateRequestHeaders(const RequestHeaders& headers) co
   }
 
   RequestHeaders::const_iterator it = headers.find(pattern[1]);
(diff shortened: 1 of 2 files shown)


jcheng5:
The constant time equals implementations I've seen generally use XOR/OR to try to ensure the compiler doesn't optimize away part of the loop. https://stackoverflow.com/a/25374036/139922

PR #199, opened 2019-02-14
Make sure that pointer refers to allocated memory

### LINKED ISSUE #198: decodeURIComponent does not handle chars longer than 22 chars
It returns a string of size 0.

``` r
devtools::install_github("rstudio/httpuv")
#> Skipping install of 'httpuv' from a github remote, the SHA1 (12744e32) has not changed since last install.
#>   Use `force = TRUE` to force installation

# 22 chars long
httpuv::decodeURIComponent("1234567890123456789012")
#> [1] "1234567890123456789012"

# 23 chars long
httpuv::decodeURIComponent("12345678901234567890123")
#> [1] ""
```

<sup>Created on 2019-02-14 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

wch:
This fixes #198.

`decodeURI` and `decodeURIComponent` now consistently return the same (correct) value.

```R
table(httpuv::decodeURI(rep("12345678901234567890123", 10000)))
#> 
#> 12345678901234567890123 
#>                   10000
table(httpuv::decodeURIComponent(rep("12345678901234567890123", 10000)))
#> 
#> 12345678901234567890123 
#>                   10000
```

### DIFF
--- a/src/httpuv.cpp
+++ b/src/httpuv.cpp
@@ -628,8 +628,7 @@ Rcpp::CharacterVector decodeURI(Rcpp::CharacterVector value) {
   for (int i = 0; i < value.size(); i++) {
     if (value[i] != NA_STRING) {
       std::string decoded = doDecodeURI(Rcpp::as<std::string>(value[i]), false);
-      const char* s = decoded.c_str();
-      out[i] = Rf_mkCharCE(s, CE_UTF8);
+      out[i] = Rf_mkCharLenCE(decoded.c_str(), decoded.length(), CE_UTF8);
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/httpuv.cpp
```diff
@@ -627,7 +627,8 @@ Rcpp::CharacterVector decodeURI(Rcpp::CharacterVector value) {
 
   for (int i = 0; i < value.size(); i++) {
     if (value[i] != NA_STRING) {
-      const char* s = doDecodeURI(Rcpp::as<std::string>(value[i]), false).c_str();
+      std::string decoded = doDecodeURI(Rcpp::as<std::string>(value[i]), false);
+      const char* s = decoded.c_str();
       out[i] = Rf_mkCharCE(s, CE_UTF8);
```
kevinushey: I'd recommend using `Rf_mkCharLenCE(decoded.c_str(), decoded.size(), CE_UTF8)` here (to avoid forcing R to scan for the null terminator in the `const char*` string)

kevinushey:
LGTM (code is correct) but would suggest using `Rf_mkCharLenCE()`.

PR #210, opened 2019-03-26
Fixes for RHEL/Centos 6

wch:
This PR has changes to make it compile on RHEL/Centos 6, which doesn't ship with a C++11 compiler.

***** 

Dockerfile for testing:

```Dockerfile
# docker build -t centos6r .

FROM centos:centos6

RUN yum -y install epel-release 
RUN yum -y update 

RUN echo "LANG=en_US.utf8" >> /etc/locale.conf
RUN localedef -c -f UTF-8 -i en_US en_US.UTF-8
RUN export LC_ALL=en_US.UTF-8

RUN yum -y install libtool R-core R-core-devel R-java R-cpp openssl-devel libcurl-devel \
  && echo 'options(repos = c(CRAN = "https://cran.rstudio.com/"), download.file.method = "libcurl")' >> /usr/lib64/R/etc/Rprofile.site

RUN echo "MAKEFLAGS=-j4" >> /usr/lib64/R/etc/Renviron
# Create directories for documentation
RUN mkdir -p /usr/share/doc/R-3.4.3/html
RUN cp /usr/lib64/R/library/stats/html/R.css /usr/share/doc/R-3.4.3/html/

# First install devtools and CRAN version httpuv so that we have dependencies
RUN R -e "install.packages(c('devtools', 'httpuv'))"

CMD bash
```

In the docker container:

```
R
devtools::install_github('rsrtudio/httpuv@fix-centos6')
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+httpuv 1.5.0.9000
+============
+
+* Fixed issues for compilers that didn't support C++11, notably on RHEL and Centos 6. ([#210](https://github.com/rstudio/httpuv/pull/210))
+
 httpuv 1.5.0
 ============
 
--- a/src/filedatasource.h
+++ b/src/filedatasource.h
@@ -21,7 +21,7 @@ class FileDataSource : public DataSource {
   int _fd;
   off_t _length;
 #endif
-  std::string _lastErrorMessage = "";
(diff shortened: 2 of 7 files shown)


kevinushey:
LGTM!

PR #215, opened 2019-04-09
Look for Upgrade header instead of 'Connection: Upgrade'

wch:
This is related to rstudio/shiny#2372. If we look for the `Upgrade` header instead of `Connection: Upgrade`, it is more robust with incorrectly-configured proxies.


wch:
I just bumped the version and updated NEWS.... on master. Oops.

PR #219, opened 2019-06-05
Consider frame complete when it has no payload

alandipert:
## Background

One day we were recording a session with [shinyloadtest](https://github.com/rstudio/shinyloadtest) and we noticed that closing the browser tab did not immediately terminate the session. In normal operation, when shinyloadtest detects that the browser's websocket has closed (via httpuv WebSocket`ws$onClose()`), it closes the "upstream" (Shiny app) websocket and records a `WS_CLOSE` message in the recording log. Then `record_session()` returns in the R console.

Instead, `record_session()` was behaving like the browser tab had not been closed.

We only observed this problem with Firefox, and only when the target application was RSC (and not e.g. `runExample("01_hello")` locally). When we tried to reproduce with Chrome, `record_session()` returned immediately after tab close, as we expected it to.

Using Firefox, `record_session()` would *eventually* return, about 20 seconds after tab close.

We couldn't expect users to wait that long, and we didn't want to encourage hitting Ctrl-c in the R console because that method introduces other potential difficulties.

### Chrome vs Firefox

We compared Chrome and Firefox with Wireshark and determined that both browsers *were* sending WebSocket Close messages.

Closing the tab in Firefox resulted in the 20 second wait, but exiting Firefox completely would cause `$onClose()` to fire.

We concluded that Chrome's WebSocket was closing immediately (at the TCP level) and Firefox was not &mdash; WebSockets in Firefox appear to live for 20 seconds after the tab in which they were created is closed.

### RSC vs local

We noticed that recording a local Shiny app worked, but recording a remote RSC app did not. Comparing traffic with Wireshark, we noticed the Close message sent in the local app case had a reason payload, `1001 - "Going Away"`. The Close message sent in the RSC app case had no payload.

### sock.js

The primary difference between recording a local app and an RSC/SSP app is the presence of [shiny-server-client](https://github.com/rstudio/shiny-server-client) and [sock.js](https://github.com/sockjs/sockjs-client) in the networking stack.

We concluded that the discrepancy in the Close message payload was accounted for by the fact that when sock.js is involved, it calls `close()` on the JS WebSocket, **not** the browser.

The version of sock.js we used registers for the browser's `onunload` event and closes the WebSocket without a close message: https://github.com/sockjs/sockjs-client/blob/0c70698bddcfab826c7b241ed709f69b5b0d41f7/lib/trans-websocket.js#L31

When the browser closes the WebSocket (in the local app case), it does so with the `1001 - "Going Away"` payload.

## Fix

Firefox's behavior (waiting 20 seconds to close the connection at the TCP layer) combined with attempting to record an app involving sock.js exposed a bug in httpuv.

Currently, `onFrameComplete()` is only invoked when a WebSocket message contains a payload, after that payload has been consumed.

With this change, `onFrameComplete()` is also invoked after the header is parsed and there is no payload.

## Testing/QA notes

Joe devised a test (comment on this PR) that checks `ws$onClose()` is called when the client closes and provides no payload. However, the test relies on the websocket package, which is not yet on CRAN.

Once websocket is on CRAN, we should add the test to our suite, and include websocket in the `Suggests` part of our `DESCRIPTION`.

### DIFF
--- a/tests/testthat/test-frame-completion.R
+++ b/tests/testthat/test-frame-completion.R
@@ -10,9 +10,9 @@ test_that("a close message with no payload is processed", {
 
   srv <- startServer("127.0.0.1", random_port, list(
     onWSOpen = function(ws) {
-      open_time <- Sys.time()
+      open_time <- as.numeric(Sys.time())
       ws$onClose(function(e) {
-        elapsed <<- Sys.time() - open_time
+        elapsed <<- as.numeric(Sys.time()) - open_time
         stopServer(srv)
       })
     }
@@ -29,11 +29,12 @@ test_that("a close message with no payload is processed", {
     })
   }
 
-  loop_start <- Sys.time()
+  loop_start <- as.numeric(Sys.time())
   while (!later::loop_empty()) {
-    if ((Sys.time() - loop_start) > as.difftime(10, units = "secs")) stop("run loop timed out")
+    loop_elapsed <- as.numeric(Sys.time()) - loop_start
+    if (loop_elapsed > 10) stop("run loop timed out")
     later::run_now()
   }
 
-  testthat::expect_true(elapsed < as.difftime(1, units = "secs"))
+  expect_true(elapsed < 1)
 })
--- a/tests/testthat/test-frame-completion.R
+++ b/tests/testthat/test-frame-completion.R
@@ -3,31 +3,37 @@
 
 context("frame completion")
 
-elapsed <- NULL
+test_that("a close message with no payload is processed", {
+  elapsed <- NULL
 
-random_port <- random_open_port()
+  random_port <- random_open_port()
 
-srv <- startServer("127.0.0.1", random_port, list(
-  onWSOpen = function(ws) {
-    open_time <- Sys.time()
-    ws$onClose(function(e) {
-      elapsed <<- Sys.time() - open_time
-      stopServer(srv)
+  srv <- startServer("127.0.0.1", random_port, list(
+    onWSOpen = function(ws) {
+      open_time <- Sys.time()
+      ws$onClose(function(e) {
+        elapsed <<- Sys.time() - open_time
+        stopServer(srv)
+      })
+    }
+  ))
+
+  on.exit(srv$stop())
+
+  # "Unnecessary" braces here to prevent `later` from attempting to
+  # run callbacks if this test is pasted at the console
+  {
+    ws_client <- websocket::WebSocket$new(sprintf("ws://127.0.0.1:%s", random_port))
+    ws_client$onOpen(function(event) {
+      ws_client$close(NA)
     })
   }
-))
 
-# "Unnecessary" braces here to prevent `later` from attempting to
-# run callbacks if this test is pasted at the console
-{
-  ws_client <- websocket::WebSocket$new(sprintf("ws://127.0.0.1:%s", random_port))
-  ws_client$onOpen(function(event) {
-    ws_client$close(NA)
-  })
-}
+  loop_start <- Sys.time()
+  while (!later::loop_empty()) {
+    if ((Sys.time() - loop_start) > as.difftime(10, units = "secs")) stop("run loop timed out")
+    later::run_now()
+  }
 
-while (!later::loop_empty()) {
-  later::run_now()
-}
-
-testthat::expect_true(elapsed < as.difftime(0.2, units = "secs"))
+  testthat::expect_true(elapsed < as.difftime(1, units = "secs"))
+})
--- a/tests/testthat/test-frame-completion.R
+++ b/tests/testthat/test-frame-completion.R
@@ -25,7 +25,10 @@ test_that("a close message with no payload is processed", {
   {
     ws_client <- websocket::WebSocket$new(sprintf("ws://127.0.0.1:%s", random_port))
     ws_client$onOpen(function(event) {
-      ws_client$close(NA)
+      # NOTE: Depends on websocketpp internals.
+      # 0 below corresponds to close::status::blank, here:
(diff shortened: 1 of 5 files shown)


jcheng5:
Something like this could serve as a unit test:

```r
elapsed <- NULL

srv <- httpuv::startServer("127.0.0.1", 8101, list(
  onWSOpen = function(ws) {
    open_time <- Sys.time()
    ws$onClose(function(e) {
      elapsed <<- Sys.time() - open_time
      httpuv::stopServer(srv)
    })
  }
))

# "Unnecessary" braces here to prevent `later` from attempting to
# run callbacks if this test is pasted at the console
{
  ws_client <- websocket::WebSocket$new("ws://127.0.0.1:8101")
  ws_client$onOpen(function(event) {
    ws_client$close(NA)
  })
}

while (!later::loop_empty()) {
  later::run_now()
}

testthat::expect_true(elapsed < as.difftime(0.2, units = "secs"))
```

tests/testthat/test-frame-completion.R
```diff
@@ -0,0 +1,33 @@
+# Regression test of
+# https://github.com/rstudio/httpuv/pull/219
+
+context("frame completion")
+
+elapsed <- NULL
```
wch: All this code from line 6 to the end should be wrapped in something like:

```R
test_that("a description", {

  # code

})

File: tests/testthat/test-frame-completion.R [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+
+random_port <- random_open_port()
+
+srv <- startServer("127.0.0.1", random_port, list(
+  onWSOpen = function(ws) {
+    open_time <- Sys.time()
+    ws$onClose(function(e) {
+      elapsed <<- Sys.time() - open_time
+      stopServer(srv)
```
wch: Instead of stopping the server in this `onClose`, I think it would be safer to use this:

```R
on.exit(srv$stop())
```

The `on.exit` executes when the `test_that()` expression is finished.

(I also used `srv$stop()` instead of `stopServer(srv)` but they're equivalent.)

wch: You could also tell it to stop from both places, in case you want to test from the console.

tests/testthat/test-frame-completion.R
```diff
@@ -0,0 +1,33 @@
...
+# "Unnecessary" braces here to prevent `later` from attempting to
+# run callbacks if this test is pasted at the console
+{
+  ws_client <- websocket::WebSocket$new(sprintf("ws://127.0.0.1:%s", random_port))
+  ws_client$onOpen(function(event) {
+    ws_client$close(NA)
+  })
+}
+
+while (!later::loop_empty()) {
```
wch: For safety, I think there should also be a timeout check for this, so that it can't run forever.

tests/testthat/test-frame-completion.R
```diff
@@ -0,0 +1,33 @@
...
+  later::run_now()
+}
+
+testthat::expect_true(elapsed < as.difftime(0.2, units = "secs"))
```
wch: No need for `testthat::` here.

Also, are you sure 0.2 seconds enough? I worry that on Travis it might fail just because the machines are so bogged down.

tests/testthat/test-frame-completion.R
```diff
@@ -0,0 +1,33 @@
...
+
+context("frame completion")
+
+elapsed <- NULL
+
+random_port <- random_open_port()
+
+srv <- startServer("127.0.0.1", random_port, list(
+  onWSOpen = function(ws) {
+    open_time <- Sys.time()
```
wch: For recording timings, I usually use `as.numeric(Sys.time())`, which converts to seconds. Then when you do the test below, instead of this:

```R
elapsed < as.difftime(0.2, units = "secs"))
```

You can do this:

```
elapsed < 0.2
```

tests/testthat/test-frame-completion.R
alandipert: It appears to work, but maybe it's not enough. What would you make it?

tests/testthat/test-frame-completion.R
```diff
@@ -0,0 +1,40 @@
...
+  ))
+
+  on.exit(srv$stop())
+
+  # "Unnecessary" braces here to prevent `later` from attempting to
+  # run callbacks if this test is pasted at the console
+  {
+    ws_client <- websocket::WebSocket$new(sprintf("ws://127.0.0.1:%s", random_port))
+    ws_client$onOpen(function(event) {
+      ws_client$close(NA)
```
wch: Quick question: what exactly does the `NA` here translate to in the C++ code? I think it's the `code` this function, right?

https://github.com/rstudio/websocket/blob/9ae1fccfbbed2978608cc4eb360db11368969674/src/websocket.cpp#L286-L303

alandipert: That looks right. Here's what I think happens:

1. NA is coerced to 0
1. 0 is the value of `close::status::blank` https://github.com/rstudio/websocket/blob/9ae1fccfbbed2978608cc4eb360db11368969674/src/lib/websocketpp/close.hpp#L52 
1. `code != close::status::blank` branch is skipped https://github.com/rstudio/websocket/blob/master/src/lib/websocketpp/impl/connection_impl.hpp#L2082

I tested with `ws_client$close(0)` and that works too. Either way seems to depend on WS internals via `close::status::blank`, and I don't know if they have an official API for closing with no code.

wch: Thanks for the explanation. Can we go with `0` here then, so that it's a little clearer what's going on?

alandipert: Surely, I changed it

PR #227, opened 2019-07-10
Fix static serving on Windows for non-ASCII paths

### LINKED ISSUE #224: Static file serving doesn't work on Windows when non-ASCII chars are in path
Example:

```R
dir.create('fü')
writeLines("Hello world!", 'fü/index.html')

dir.create('foo')
writeLines("Hello world!", 'foo/index.html')

library(httpuv)


  list(
    call = function(req) {},
    staticPaths = list(
      "/bad" = "fü",
      "/good" = "foo"
    )
  )
)

# s$stop()
```

After running this code, visit http://localhost:5000/bad and http://localhost:5000/good. The first one maps to fü, and it will give a 404. The second one maps to foo, and it works correctly.

wch:
This fixes #224.

### DIFF
--- a/src/winutils.cpp
+++ b/src/winutils.cpp
@@ -1,0 +1,56 @@
+#include "winutils.h"
+#ifdef _WIN32
+
+#include <windows.h>
+#include <vector>
+
+std::string wideToUtf8(const std::wstring& value)
+{
+   if (value.size() == 0)
+      return std::string();
+
+   const wchar_t * cstr = value.c_str();
+   int chars = ::WideCharToMultiByte(CP_UTF8, 0,
+                                     cstr, -1,
+                                     nullptr, 0, nullptr, nullptr);
+   if (chars == 0) {
+      return std::string();
+   }
+
+   std::vector<char> result(chars, 0);
+   chars = ::WideCharToMultiByte(CP_UTF8,
+                                 0,
+                                 cstr,
+                                 -1,
+                                 &(result[0]),
+                                 result.size(),
+                                 nullptr, nullptr);
+
+   return std::string(&(result[0]));
+}
+
+
+std::wstring utf8ToWide(const std::string& value,
+                        const std::string& context)
+{
+   if (value.size() == 0)
+      return std::wstring();
+
+   const char * cstr = value.c_str();
+   int chars = ::MultiByteToWideChar(CP_UTF8, 0,
+                                     cstr, -1,
+                                     nullptr, 0);
+   if (chars == 0) {
+      return std::wstring();
+   }
+
+   std::vector<wchar_t> result(chars, 0);
+   chars = ::MultiByteToWideChar(CP_UTF8, 0,
+                                 cstr, -1,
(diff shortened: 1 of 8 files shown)


src/winutils.cpp
```diff
@@ -0,0 +1,56 @@
...
+                                 &(result[0]),
+                                 result.size(),
+                                 nullptr, nullptr);
+
+   return std::string(&(result[0]));
+}
+
+
+std::wstring utf8ToWide(const std::string& value,
+                        const std::string& context)
```
kevinushey: Looks like this is unused?

src/winutils.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+   if (value.size() == 0)
+      return std::wstring();
+
+   const char * cstr = value.c_str();
+   int chars = ::MultiByteToWideChar(CP_UTF8, 0,
+                                     cstr, -1,
```
kevinushey: nit: you could use `value.data()`, `value.size()` here

src/winutils.cpp
```diff
@@ -0,0 +1,56 @@
...
+      return std::wstring();
+   }
+
+   std::vector<wchar_t> result(chars, 0);
+   chars = ::MultiByteToWideChar(CP_UTF8, 0,
+                                 cstr, -1,
+                                 &(result[0]),
+                                 result.size());
+
+   return std::wstring(&(result[0]));
```
kevinushey: `std::wstring(result.begin(), result.end())`?

src/winutils.cpp
kevinushey: (or at least `value.size()`, to avoid having to scan the string for the null terminator)

wch: Should this pattern also be used below, in line 47?

And similarly, should it be used above with `WideCharToMultiByte` on lines 13 and 21?

kevinushey: Yes, I think so. (The code as-is is still correct but it's better to use the string length if you already know it)

wch: Maybe someone should fix this? :)
https://github.com/rstudio/rstudio/blob/72132fdf30e0803f840ff98fc82917c52ab930ec/src/cpp/core/Win32StringUtils.cpp#L29-L82

That's where this code is originally from.

kevinushey: Those darn IDE folks... can't trust 'em  👀

wch: I found that I needed to use `value.data(), value.size() + 1` because without the `+1`, it will not include the NULL-terminator character.

I think there's also the risk that if the string happens to have a `\0` in the middle of it, then `size()` will report a value that's longer than the length up to the `\0`. See https://stackoverflow.com/a/11752711/412655

kevinushey: You're right. Perhaps it is safer to ensure we don't construct a path with an embedded NUL.

wch: OK, I'm sticking with the existing code for this, then.

PR #234, opened 2019-08-28
Add randomPort function, and add quiet option for startServer

### LINKED ISSUE #194: Suppress the error message "createTcpServer: address already in use"
Is there a way to suppress this error message? For example, it may show up when **shiny** looks for a random port: https://github.com/rstudio/shiny/blob/8ae31eb/R/server.R#L742 but the error message is not really useful to users.

### LINKED ISSUE #233: Have err_printf respect log_level
Currently if I want to use something like `servr:::port_available`, on a failure I will always get an error message from `createTcpServer`:

```r
s = httpuv::startServer("0.0.0.0", 5000, list())
servr:::port_available(5000L, "0.0.0.0")
## createTcpServer: address already in use
## [1] FALSE
```

in a function like this, this kind of failure is expected so it would be nice if it were possible to suppress these messages temporarily via `logLevel("OFF")`.

wch:
Closes #194, closes #233. The `quiet` option suppresses messages that are normally printed to console (and can't be intercepted with `capture.output`).

Update: This PR now adds a new function, `randomPort`.

Testing notes. The code below should have the behavior as described.

```R
port <- 5001
# Start server
s <- startServer("127.0.0.1", port, list())


# Should print out "createTcpServer: address already in use", and throw an error
startServer("127.0.0.1", port, list())

# Should throw error only (and not print "createTcpServer: address already in use")
startServer("127.0.0.1", port, list(), quiet = TRUE)

# Should print "createTcpServer: address already in use" and return NULL
tryCatch(
  startServer("127.0.0.1", port, list()),
  error = function(e) {}
)

# Should just print NULL
tryCatch(
  startServer("127.0.0.1", port, list(), quiet = TRUE),
  error = function(e) {}
)



# Should return a random open port
randomPort()

# Should return 5002 (because 5001 is already taken)
randomPort(min = port, max = port + 1)
```


jcheng5:
@wch Did you want this in the upcoming release of httpuv? Or the next one?

wch:
I was hoping to get it in this release. Actually, it might make sense to add in some code that finds an available port, as described in rstudio/shiny#2562.

R/random_port.R
```diff
@@ -0,0 +1,117 @@
...
+#'
+#' Finds a random available TCP port for listening on, within a specified range
+#' of ports. The default range of ports to check is 1024 to 49151, which is the
+#' set of TCP User Ports. This function automatically excludes some ports which
+#' are considered unsafe by web browsers.
+#'
+#' @inheritParams runServer
+#' @param min Minimum port number.
+#' @param max Maximum port number.
+#' @param n Number of ports to try before giving up.
```
cpsievert: Missing `@param host`

wch: The `host` documentation is inherited from `runServer`.

PR #237, opened 2019-08-29
Deal gracefully with apps that don't have $call(), also unit tests

jcheng5:
(Might want to wait until post-1.5.2 for this one.)

If you have an app without a `call()` but with static paths, then any requests that don't match the static paths end up giving an error 500 with something about not being able to find `func`. This PR causes a graceful 404 to be returned in this situation instead.

Also contains unit tests for path traversal attack mitigation (which is how I discovered this in the first place).

### DIFF
--- a/tests/testthat/test-static-paths.R
+++ b/tests/testthat/test-static-paths.R
@@ -68,21 +68,6 @@ test_that("Basic static file serving", {
   expect_identical(rawToChar(r$content), "404 Not Found\n")
   expect_equal(h$`content-length`, "14")
 
-  # Bad request (400)
-  # jcheng 2019-09-28: I tried to do this same test with /1/../index.html but
-  # it seems like the actual request being made would always end up being
-  # /1/index.html instead, which would succeed.
-  r <- fetch(local_url("/1/..%2Findex.html", s$getPort()))
-  h <- parse_headers_list(r$headers)
-  expect_equal(r$status_code, 400)
-  expect_identical(rawToChar(r$content), "400 Bad Request\n")
-
-  # Missing file (404)
-  r <- fetch(local_url("/1/..\\index.html", s$getPort()))
-  h <- parse_headers_list(r$headers)
-  expect_equal(r$status_code, 404)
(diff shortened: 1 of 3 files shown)


tests/testthat/test-static-paths.R
```diff
@@ -68,6 +68,21 @@ test_that("Basic static file serving", {
   expect_identical(rawToChar(r$content), "404 Not Found\n")
   expect_equal(h$`content-length`, "14")
 
+  # Bad request (400)
+  # jcheng 2019-09-28: I tried to do this same test with /1/../index.html but
+  # it seems like the actual request being made would always end up being
+  # /1/index.html instead, which would succeed.
+  r <- fetch(local_url("/1/..%2Findex.html", s$getPort()))
```
wch: The `fetch()` function uses curl, which will automatically resolve the `..` before sending the request. If you want to send a verbatim HTTP request, you can use `http_request_con()` instead of `fetch()`. For example:

https://github.com/rstudio/httpuv/blob/2a2123f9cd6440a10ff19e3b7f4a68d7e1608c93/tests/testthat/test-static-paths.R#L568

tests/testthat/test-static-paths.R
```diff
@@ -68,6 +68,21 @@ test_that("Basic static file serving", {
...
+  h <- parse_headers_list(r$headers)
+  expect_equal(r$status_code, 400)
+  expect_identical(rawToChar(r$content), "400 Bad Request\n")
+
+  # Missing file (404)
+  r <- fetch(local_url("/1/..\\index.html", s$getPort()))
```
wch: Does curl turn this into a `%5C`?

tests/testthat/test-static-paths.R
wch: I removed the `..` test because it's the same as ones below (like the one I linked to on line 568).

I moved the backslash test to its own test block, and added a few more variants on it.

tests/testthat/test-static-paths.R
wch: I found that the `\` does not get converted to a `%5C`, so I added tests for both.

Issue #247, opened 2019-10-24
For static files, if header Content-Length is 0 no HTTP 400 should be returned


LHaferkamp:
As described here: https://stackoverflow.com/a/58541479/2854141
a HTTP 400 is returned when serving static files if header `Content-Length:0` is set. This case should be considered as valid and be excluded in the check in webapplication.cpp
```

// Make sure that there's no message body.
  if (pRequest->hasHeader("Content-Length") || pRequest->hasHeader("Transfer-Encoding")) {
    return error_response(pRequest, 400);
  }
```

wch:
Here's what I see in https://tools.ietf.org/html/rfc7230#section-3.3.2:
> A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body.

If you think that this doesn't apply to the situation you are encountering, can you please provide an example of the exact HTTP request that is used? It's not entirely clear to me from reading the Stack Overflow answer.

LHaferkamp:
You are right, this RFC statement applies to the situation. Unfortunately I can't control this additional header injected by the Azure proxy. So it would be really beneficial to be less strict, and it would still comply to the  RFC as its not a MUST to decline such requests

wch:
OK, that sounds reasonable.

wch:
Closed by #248.

PR #248, opened 2019-10-29
Issue #247 fixed, Content-Length:0 is valid header for static files

LHaferkamp:


wch:
@LHaferkamp Can you address the things I mentioned in the previous message?

PR #262, opened 2020-04-06
Add a timestamp to the Rook request object.

atheriel:
This PR adds a `httpuv.timestamp` member containing a timestamp compatible with `Sys.time()` to the Rook request object. You can use these timestamps it to measure and report on request latency:

```r
httpuv::runServer(
  "127.0.0.1", 8080,
  list(call = function(req) {
    elapsed_ms <- (unclass(Sys.time()) - req$httpuv.timestamp) * 1000
    list(
      status = 200L,
      headers = list("Content-Type" = "text/plain"),
      body = sprintf("Latency (ms): %.3f\n", elapsed_ms)
    )
  })
)
```

Importantly, this timestamp is captured when the request is received at the C++ level in the background thread, which has two advantages:

1. We can much more accurately capture the total time the client actually waits for the request to get processed, especially if the R thread is congested or running GC.

2. We can instrument the time it takes before the R-level callback actually runs.

More broadly, the timestamp makes it easier to log measures of request latency in `httpuv`-based applications, and reduces the overhead of common ways to do so -- e.g. using Plumber hooks.

I'm unsure about whether `httpuv.timestamp` is a good name, and am happy to take suggestions for improvements.

Also, I have little faith in my C++ (though I've made sure to use C++11-compatible interfaces here).

Finally, if you're interested in R's internal `Sys.time()` implementation, the source is in `time.c:currentTime()`.


atheriel:
CI failures seem unrelated.

wch:
The CI failures are due to https://github.com/RcppCore/Rcpp/issues/1048.

atheriel:
Tweaked slightly and rebased to fix merge conflicts.

PR #263, opened 2020-04-19
Support platform IBM i (OS400)

dmabupt:
On IBM i, we need to link to the shared library libuv.so, rather than compiling it from the source.

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -20,6 +20,9 @@ endif
 ifeq ($(UNAME), OpenBSD)
 PKG_LIBS += -lkvm
 endif
+ifeq ($(UNAME), OS400)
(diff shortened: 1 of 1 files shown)


jcheng5:
Thanks... does it fail to compile? Or do you just prefer the .so?

dmabupt:
> Thanks... does it fail to compile? Or do you just prefer the .so?

The version (1.27.0) of the build configuration is not completely compatible with IBM i. It will fail to compile libuv on IBM i.
So we may upgrade the integrated libuv to a newer version (1.36.0 or above), or link to the latest shared libuv library.

wch:
Update: httpuv 1.5.3.1 is now on CRAN, and it uses libuv 1.37.0.

PR #277, opened 2020-06-01
Fix header parsing for values that span more than one call to _on_header_value

### LINKED ISSUE #275: Large request headers can be truncated
When request headers are large, they can get truncated. 

Example:

```R
library(httpuv)

# Create a web server app which returns the value of "test-header".
handle_headers <- function(req) {
  str(as.list(req))
  message("headers")
}
handle_request <- function(req) {

  # Store the headers in a global var named z, for inspection later.
  z <<- req$HEADERS

  list(
    status = 200L,
    headers = list('Content-Type' = 'text/plain'),
    # Use paste0("", ...) in case the header is NULL
    body = paste0("", req$HTTP_TEST_HEADER)
  )
}
s <- httpuv::startServer("0.0.0.0", 5000,
  list(
    onHeaders = function(req) handle_headers(req),
    call = function(req) handle_request(req)
  )
)


# Request a URL from the local web app
fetch <- function(url, headers = list()) {
  h <- curl::new_handle()
  do.call(curl::handle_setheaders, c(list(h), headers))

  # Make the request
  x <- NULL
  curl::curl_fetch_multi(
    url,
    handle = h,
    done = function(res) { x <<- res },
    fail = function(msg) {
      x <<- FALSE
      message("Failed: ", msg)
    }
  )
  # Pump both the curl and httpuv/later event loops until the download is finished.
  while (is.null(x)) {
    curl::multi_run(0)
    later::run_now()
  }

  x
}

# ===============================
# Create a request with a large header (81000 bytes)
long_string <- paste0(rep(".", 81000), collapse = "")
x <- fetch("http://127.0.0.1:5000",  list("test-header" = long_string))
content <- rawToChar(x$content)
nchar(content)
#> [1] 65380
identical(content, long_string)
#> [1] FALSE
```

The headers are stored in `z`. If you inspect them, there's a header named `""` which has the remaining part of the header. Adding up the sizes of `"test-header"` and `""`, we get the original header size, which is 81000.

```R
names(z)
#> [1] ""                "accept"          "accept-encoding" "host"           
#> [5] "test-header"     "user-agent"     
nchar(z)
#>                       accept accept-encoding            host     test-header 
#>        15620               3              13              14           65380 
#>   user-agent 
#>           46 

nchar(z[1]) + nchar(z[5])
#> 81000 
```

@aronatkins also provided a Python test app, which splits up the HTTP request into small TCP messages.

<details>
<summary>Details</summary>

```python
#!/usr/bin/env python3
# Usage:
# ./test.py 127.0.0.1 5000

import json
import socket
import sys

host,port=(sys.argv[1], int(sys.argv[2]))
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.connect((host,port))

groups = []
for i in range(1000):
    groups.append("group-%d" % i)

value = json.dumps({'user':'admin','groups':groups})

message_template = """GET / HTTP/1.1
Host: %s:%s
Test-Header: %s

"""

message = message_template % (
    host,
    port,
    value,
)
message = message.encode()

sock = socket.socket()
sock.connect( (host, int(port)) )

sendsize=1
cur=0
while cur < len(message):
    sent = sock.send(message[cur:cur+sendsize])
    if sent == 0:
        raise RuntimeError("send sent nothing")
    cur = cur + sent

chunks = []
while True:
    chunk = sock.recv(1024)
    if chunk:
        chunks.append(chunk)
    else:
        break
print(''.join(chunks))

```

</details>

wch:
This fixes #275.

cc @aronatkins

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Package: httpuv
 Type: Package
 Encoding: UTF-8
 Title: HTTP and WebSocket Server Library
-Version: 1.5.3.9000
+Version: 1.5.3.9001
 Authors@R: c(
     person("Joe", "Cheng", role = c("aut"), email = "joe@rstudio.com"),
     person("Winston", "Chang", role = c("aut", "cre"), email = "winston@rstudio.com"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 httpuv 1.5.3.9000
 ==============
 
+* Fixed #275: Large HTTP request headers could get truncated if they spanned
+  more than one TCP message. (#277)
+
 * Fixed build for Solaris. (#271)
 
 * Fixed a test that had incorrect logic. (#272)
--- a/src/httprequest.cpp
+++ b/src/httprequest.cpp
@@ -138,8 +138,7 @@ void HttpRequest::_newRequest() {
   _handling_request = true;
   _headers.clear();
   _response_scheduled = false;
-
-  _last_header_state = "start";
+  _last_header_state = START;
 
   // Schedule on main thread:
   //   this->_initializeEnv();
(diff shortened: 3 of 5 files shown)


wch:
The fields already had the same append treatment -- the `std::copy` with the `std::back_inserter` appends incoming data to `_lastHeaderField`. 

https://github.com/rstudio/httpuv/blob/7243a5703d51c94ec9e578ba2188381e97f02873/src/httprequest.cpp#L234

(`_lastHeaderField` is now cleared when `_last_header_state` transitions to `FIELD`.)

I've also added a unit test for this.

PR #284, opened 2020-09-07
Fix sha1 calculation on big-endian systems.

QuLogic:
Even though there is a check for `WORDS_BIGENDIAN`, this does nothing because that is `#define`d by R headers, which are not included by this code. But instead of fixing the `#define`, it is simpler to load the `block` workspace from the `buffer` as big-endian `uint32_t` directly.

This does not appear exercised by tests, because nothing appears to read `Sec-WebSocket-Accept` anywhere that I can tell. However, if you run the latest version `websocket`'s tests (which use `httpuv` as a server), then they fail on big-endian machines because `websocket` _does_ calculate the sha1 correctly and verify `Sec-WebSocket-Accept`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 httpuv 1.5.4.9000
 =================
 
+* Fix SHA1 calculation, and thus WebSocket server handshakes, on big-endian systems.
 
 httpuv 1.5.4
 ============
--- a/src/sha1/sha1.c
+++ b/src/sha1/sha1.c
@@ -86,6 +86,9 @@ A million repetitions of "a"
 #include <string.h>
 #include <stdint.h>
 
+/* To get WORDS_BIGENDIAN definition. */
+#include <Rconfig.h>
+
 #include "sha1.h"
 
 void SHA1_Transform(uint32_t state[5], const uint8_t buffer[64]);


wch:
What platform are you using?

I'm wary of taking a change like this, where it's very difficult for us to verify the correctness of the code. Is there another version of this SHA1 C library that handles endianness without that define?

QuLogic:
Well, that's why I removed the dependency on the `WORDS_ENDIAN` define entirely. The new code reads the bytes as big-endian directly, which is much easier to verify than the bit-twiddling, and is endian-neutral. The change looks larger than that because I dropped the `union`, which is no longer necessary. I can split that into separate commits if it's too large.

I've tested this against websocket 1.3.1 on both little-endian (x86*, arm*, ppc64le) and big-endian (s390x) Fedora.

wch:
The SHA1 library is a third-party library, and is somewhat widely used (it's apparently used by Chromium). I'd much rather go with an approach that minimizes changes to the existing library, for reasons of long-term maintenance. The code you've written may be simpler for you to understand, but it is not simpler for us to verify and maintain.

Can you just add `#include <Rconfig.h>` to the top (with a comment about why it's there)? I believe it should get the `WORDS_BIGENDIAN` macro from there.

QuLogic:
Oh, I thought you wrote the FIXME when changing to WORDS_BIGENDIAN, but I see that it's in the Chromium source as well.

That would work also.

wch:
Thanks!

PR #286, opened 2020-10-29
Update ReadMe

vikram-rawat:
The Old link is dead and this is the new link mentioned on the old link.


schloerke:
Thanks for the link update and sorry it sat for so long. The README link has since been updated to libuv/libuv on main (bd0fe48), so this is superseded. Fixed that and a few other stale links in #444, with credit to you in NEWS. Closing this PR.

Issue #195, opened 2019-02-12
Exception occurs when reponse headers is not specified, NULL, or an empty list


wch:
In the example app below, the response is `An exception occurred.` when `headers` is unset, `NULL`, or an empty list. However, it does work if it is an empty _named_ list.

```R
library(httpuv)
s <- startServer("127.0.0.1", 5000,
  list(
    call = function(req) {
      if (req$PATH_INFO == "/unset") {
        list(
          status = 200,
          body = "200 OK\n"
        )
      } else if (req$PATH_INFO == "/null") {
        list(
          status = 200,
          headers = NULL,
          body = "200 OK\n"
        )
      } else if (req$PATH_INFO == "/empty_list") {
        list(
          status = 200,
          headers = list(),
          body = "200 OK\n"
        )
      } else if (req$PATH_INFO == "/empty_named_list") {  # This one works
        list(
          status = 200,
          headers = list(a=1)[0],
          body = "200 OK\n"
        )
      }
    }
  )
)
```

I believe it's because this line of code assumes it's a named list: https://github.com/rstudio/httpuv/blob/8c48fea/src/webapplication.cpp#L235

PR #289, opened 2020-11-26
Allow headers to be NULL or an unnamed list

wch:
Previously, `headers` had to be a named list. In order to have no headers (besides the ones httpuv adds automatically), `headers` would have to be an empty _named_ list, which is a little awkward to create. With this change, `headers` can be `NULL`, `list()` (an empty unnamed list), or it can be unset.

For example, this app would not work before, but it does now:

```R
s <- startServer("127.0.0.1", 5000,
  list(
    call = function(req) {
      list(
        status = 200L,
        headers = list(),
        body = ""
      )
    }
  )
)
```

The result:

```
$ curl -i http://localhost:5000
HTTP/1.1 200 OK
Date: Thu, 26 Nov 2020 16:33:55 GMT
Content-Length: 0

```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@ httpuv 1.5.4.9000
 
 * Fix SHA1 calculation, and thus WebSocket server handshakes, on big-endian systems. (#284)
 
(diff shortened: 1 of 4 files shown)


atheriel:
This fixes #195.

wch:
Thanks, I forgot about that issue. I've updated the NEWS to refer to it.

PR #290, opened 2020-11-27
Add partial support for the 'Range' header when serving static files

### LINKED ISSUE #259: support for range requests
httpuv's static file serving doesn't currently support range requests (i.e. requests for only part of a file). This would be very useful! `servr` supports this, but by reading the static content into memory first rather than reading from disk (see https://github.com/yihui/servr/issues/35 and https://github.com/yihui/servr/pull/41).
The need for range request support in httpuv/shiny has been mentioned in various places, see e.g.:

- https://github.com/rstudio/shiny/pull/747 (but that PR not merged, and I don't think it will work now anyway because static file handling is now deferred by default to httpuv, not handled by shiny itself)
- https://community.rstudio.com/t/range-requests-in-pdf-js-used-in-r-shiny/11488
- https://stackoverflow.com/questions/49624699/adding-http-headers-in-shiny-to-play-audio
- emuR uses something very similar to servr's implementation: https://github.com/IPS-LMU/emuR/blob/master/R/emuR-server.R#L247

Range request support in shiny should (?) be relatively straightforward once httpuv supports it, just requiring a couple of extra headers to be passed to the shiny app.

atheriel:
As proposed in #259. I haven't done extensive 'real-world' testing, but I've been able to use this to enable rewinding/skipping in HTML `<audio>` tags in a Shiny app.

This PR includes

* Parsing and validation of the [most common forms](https://tools.ietf.org/html/rfc7233#section-3.1) of the `Range` header, namely `bytes=0-1000` and `bytes=1001-`, for serving slices of files and enabling resumable downloads, etc.

* Support for returning partial files matching these slices on Unix and Windows platforms.

* Support for returning the appropriate `Content-Range` header and HTTP 206 responses for these requests.

* Tests for these features.

It does not support Windows as of yet, largely because I don't have a good test environment -- it should be possible to accomplish by passing some [`OVERLAPPED` structure](https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped) to `ReadFile()`.

There is also no support for advanced features such as [multipart ranges](https://tools.ietf.org/html/rfc7233#appendix-A) or [the `If-Range` header](https://tools.ietf.org/html/rfc7233#section-3.2).

Since `Range` header support is always optional (a server can just respond with the whole file and a HTTP 200 instead), we just fall back on existing behaviour in these cases instead of issuing some kind of error.

~Along these same lines, we don't yet advertise that we support the `Range` header by sending an [`Accept-Ranges: bytes` header](https://tools.ietf.org/html/rfc7233#section-2.3) on other GET/HEAD requests.~

Closes #259.

### DIFF
--- a/src/filedatasource-unix.cpp
+++ b/src/filedatasource-unix.cpp
@@ -35,7 +35,8 @@ FileDataSourceResult FileDataSource::initialize(const std::string& path, bool ow
       return FDS_ISDIR;
     }
 
-    _length = info.st_size;
+    _payloadSize = info.st_size;
+    _fsize = info.st_size;
 
     if (owned && unlink(path.c_str())) {
       // Print this (on either main or background thread), since we're not
@@ -49,7 +50,24 @@ FileDataSourceResult FileDataSource::initialize(const std::string& path, bool ow
 }
 
 uint64_t FileDataSource::size() const {
-  return _length;
+  return _payloadSize;
+}
+
+uint64_t FileDataSource::fileSize() const {
+  return _fsize;
+}
+
+bool FileDataSource::setRange(uint64_t start, uint64_t end) {
+  ASSERT_BACKGROUND_THREAD()
+  if (end > _fsize || end < start) {
+    return false;
+  }
+  if (lseek(_fd, start, SEEK_SET) < 0) {
+    err_printf("Error in lseek: %d\n", errno);
+    return false;
+  }
+  _payloadSize = end - start + 1;
+  return true;
 }
 
 uv_buf_t FileDataSource::getData(size_t bytesDesired) {
--- a/src/filedatasource-win.cpp
+++ b/src/filedatasource-win.cpp
@@ -49,18 +49,39 @@ FileDataSourceResult FileDataSource::initialize(const std::string& path, bool ow
   }
 
 
-  if (!GetFileSizeEx(_hFile, &_length)) {
+  if (!GetFileSizeEx(_hFile, &_fsize)) {
     CloseHandle(_hFile);
     _hFile = INVALID_HANDLE_VALUE;
     _lastErrorMessage = "Error retrieving file size for " + path + ": " + toString(GetLastError()) + "\n";
     return FDS_ERROR;
   }
+  _payloadSize = _fsize.QuadPart;
 
   return FDS_OK;
 }
 
 uint64_t FileDataSource::size() const {
-  return _length.QuadPart;
+  return _payloadSize;
+}
+
+uint64_t FileDataSource::fileSize() const {
+  return _fsize.QuadPart;
+}
+
+bool FileDataSource::setRange(uint64_t start, uint64_t end) {
+  ASSERT_BACKGROUND_THREAD()
+  if (end > _fsize.QuadPart || end < start) {
+    return false;
+  }
+  LARGE_INTEGER position = {0};
+  position.QuadPart = start;
+  // When using FILE_BEGIN, position is interpreted as an unsigned value.
+  if (!SetFilePointerEx(_hFile, position, NULL, FILE_BEGIN)) {
+    err_printf("Error in SetFilePointerEx: %d\n", GetLastError());
+    return false;
+  }
+  _payloadSize = _fsize.QuadPart - start + 1;
+  return true;
 }
 
 uv_buf_t FileDataSource::getData(size_t bytesDesired) {
--- a/src/filedatasource.h
+++ b/src/filedatasource.h
@@ -16,10 +16,12 @@ enum FileDataSourceResult {
 class FileDataSource : public DataSource {
 #ifdef _WIN32
   HANDLE _hFile;
-  LARGE_INTEGER _length;
+  uint64_t _payloadSize;
+  LARGE_INTEGER _fsize;
 #else
   int _fd;
-  off_t _length;
+  off_t _payloadSize;
+  off_t _fsize;
 #endif
   std::string _lastErrorMessage;
 
@@ -31,7 +33,12 @@ public:
   }
 
   FileDataSourceResult initialize(const std::string& path, bool owned);
+  // Total size of the data in this source, in bytes, after accounting for file
+  // size and range settings.
   uint64_t size() const;
+  // Total size of the underlying file in this source, in bytes.
+  uint64_t fileSize() const;
+  bool setRange(uint64_t start, uint64_t end);
   uv_buf_t getData(size_t bytesDesired);
   void freeData(uv_buf_t buffer);
   // Get the mtime of the file. If there's an error, return 0.
--- a/src/webapplication.cpp
+++ b/src/webapplication.cpp
@@ -563,6 +563,70 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
     }
   }
 
+  // Check the HTTP Range header, if it has been specified.
+  //
+  // We only support a limited form of the HTTP Range header (something like
+  // 'bytes=(\\d+)-(\\d+)?'), so some client requests can't be fulfilled and we
+  // need to fall back on a 200 OK instead.
+  bool hasRange = false;
+  uint64_t start = 0;
+  uint64_t end = pDataSource->fileSize() - 1;
+  if (pRequest->hasHeader("Range")) {
+    std::string rangeHeader = pRequest->getHeader("Range");
+    size_t cursor = rangeHeader.find("bytes=");
+    size_t sep = rangeHeader.find("-", cursor + 6);
+    if (cursor != std::string::npos && sep != std::string::npos && sep != cursor + 6) {
+      // At this point we can be sure we have something like 'bytes=([^-]+)-.*'
+      // and we can try parsing the range itself.
+      char *ptr = ((char *) rangeHeader.c_str()) + cursor + 6; // Skip over 'bytes='.
+      char *endptr = ptr;
+      // strtoull() will return 0 when it can't find a number or ULLONG_MAX on
+      // overflow. Since we can't have a file larger than ULLONG_MAX anyway,
+      // this is OK because we'll just return HTTP 416 below.
+      start = strtoull(ptr, &endptr, 10);
+      if (start == 0 && (endptr == ptr || *endptr != '-')) {
+        // Either there was no number at all or there was garbage *after* the
+        // number but before the '-' separator. This is invalid range syntax
+        // from the client.
+        return error_response(pRequest, 400);
+      }
+      if (start >= pDataSource->fileSize()) {
+        boost::shared_ptr<HttpResponse> pResponse = error_response(pRequest, 416);
+        pResponse->headers().push_back(
+          std::make_pair("Content-Range", "bytes */" + toString(pDataSource->fileSize()))
+        );
+        return pResponse;
+      }
+      ptr = endptr + 1; // Skip the '-'.
+      if (*ptr != '\0') {
+        end = strtoull(ptr, &endptr, 10);
+        if (*endptr != '\0' && *endptr != ',') {
+          // We hit a non-digit, non-multirange character at some point.
+          return error_response(pRequest, 400);
+        }
+        if (end < start) {
+          return error_response(pRequest, 400);
+        }
+        ptr = endptr;
+      }
+      // A comma indicates we're parsing a multipart range, which we don't
+      // support. So we need to fallback on 200 behaviour instead when we detect
+      // one.
+      if (*ptr != ',') {
+        if (end > pDataSource->fileSize() - 1) {
+          // The client might not know the size, so the range end is supposed to
+          // be redefined to be the last byte in this case instead of issuing an
+          // error. This also catches overflow in strtoull() above, which would
+          // be OK.
+          //
+          // See: https://tools.ietf.org/html/rfc7233#section-2.1
+          end = pDataSource->fileSize() - 1;
+        }
+        hasRange = pDataSource->setRange(start, end);
+      }
+    }
+  }
+
   // ==================================
   // Create the HTTP response
   // ==================================
@@ -583,6 +647,10 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
   if (client_cache_is_valid) {
     pDataSource2.reset();
     status_code = 304;
+  }
+
+  if (hasRange) {
+    status_code = 206;
   }
 
   boost::shared_ptr<HttpResponse> pResponse = boost::shared_ptr<HttpResponse>(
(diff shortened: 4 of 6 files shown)


atheriel:
Conflicts and tests on Windows fixed.

src/webapplication.cpp
```diff
@@ -578,11 +615,22 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
 
   if (method == "HEAD") {
     pDataSource2.reset();
+    hasRange = false;
```
wch, line 618: For HEAD requests, it should add the `Accept-Ranges` header, I believe, because that's how clients check if range requests are supported.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

src/webapplication.cpp
```diff
@@ -563,6 +564,42 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
...
+      }
+      if (end < start) {
+        // HTTP 416 seems to be for syntactically valid but otherwise
+        // unfulfillable requests, so use HTTP 400 for bad syntax instead.
+        return error_response(pRequest, 400);
+      }
+      if (start >= pDataSource->size()) {
+        boost::shared_ptr<HttpResponse> pResponse = error_response(pRequest, 416);
+        pResponse->headers().push_back(
+          std::make_pair("Content-Range", std::string("bytes */") + toString(pDataSource->size()))
```
wch, line 587: The 416 response uses `size()` to report the file size, but this makes me a bit uneasy, because that's not exactly what `size()` means anymore (as described above). At the very least, there should be a comment about why it happens to work here; in case there are future changes to the code that causes `size()` to change before getting here.

src/webapplication.cpp
```diff
@@ -563,6 +564,42 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
...
+      start = std::stoi(rangeMatch[1].str().c_str());
+      if (!rangeMatch[2].str().empty()) {
+        end = std::stoi(rangeMatch[2].str().c_str());
+      }
+      if (end < start) {
+        // HTTP 416 seems to be for syntactically valid but otherwise
+        // unfulfillable requests, so use HTTP 400 for bad syntax instead.
+        return error_response(pRequest, 400);
+      }
+      if (start >= pDataSource->size()) {
```
wch, line 584: What about other invalid values, like if `end` is greater than the file, or if `end<start`?

src/webapplication.cpp
```diff
@@ -563,6 +564,42 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
     }
   }
 
+  // Check the HTTP Range header, if it has been specified.
+  std::smatch rangeMatch;
+  int start = 0, end = pDataSource->size() - 1;
+  if (pRequest->hasHeader("Range")) {
```
wch, line 570: What happens in the case of a multipart range request?

src/webapplication.cpp
wch: Also, it's good that this check is for `>=`. There should also be a test for cases where the `start` is the same as the size of the file -- which should be an error because the counting starts from 0.

tests/testthat/test-static-paths.R
```diff
@@ -808,3 +808,69 @@ test_that("Paths with non-ASCII characters", {
...
+  # End of a smaller file than expected.
+  handle_setheaders(h, "Range" = "bytes=1000-2000")
+  r <- fetch(local_url("/mtcars.csv", s$getPort()), h)
+  expect_identical(r$status_code, 206L)
+  expect_identical(
+    parse_headers_list(r$headers)$`content-range`,
+    sprintf("bytes 1000-%d/%d", file_size - 1, file_size)
+  )
+  expect_equal(length(r$content), 303)
+})
```
wch, line 876: It would also be good to have tests of edge cases, like requests of length 0 and 1, as well as requests of the offsets 0, 1, the last byte, and the last byte + 1.

tests/testthat/test-static-paths.R
```diff
@@ -808,3 +808,69 @@ test_that("Paths with non-ASCII characters", {
...
+
+  # Start of a file.
+  handle_setheaders(h, "Range" = "bytes=0-499")
+  r <- fetch(local_url("/mtcars.csv", s$getPort()), h)
+  expect_identical(r$status_code, 206L)
+  expect_identical(
+    parse_headers_list(r$headers)$`content-range`,
+    sprintf("bytes 0-499/%d", file_size)
+  )
+  expect_equal(length(r$content), 500)
```
wch, line 855: These tests should also check that the correct content is returned.

src/webapplication.cpp
```diff
@@ -578,11 +615,22 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
 
   if (method == "HEAD") {
     pDataSource2.reset();
+    hasRange = false;
   }
 
   if (client_cache_is_valid) {
     pDataSource2.reset();
     status_code = 304;
+    hasRange = false;
```
wch, line 624: I'd prefer not setting this variable here (and also in the previous `if` statement). If possible I'd like to avoid heading down a path where all sorts of logic gets intertwined, because that will make it harder to maintain and debug in the future. It may be reasonably manageable now, but if we want to handle more special logic, it could get difficult.

wch (changes requested):

src/webapplication.cpp
atheriel: It's ignored and we fall back on the usual `200` behaviour. Since `Range` support is optional, clients should always be equipped to have their request ignored.

I can add a unit test making this explicit.

src/filedatasource-unix.cpp
```diff
@@ -52,6 +52,18 @@ uint64_t FileDataSource::size() const {
   return _length;
 }
 
+bool FileDataSource::setRange(size_t start, size_t end) {
+  if (end > _fsize) {
+    return false;
+  }
+  if (lseek(_fd, start, SEEK_SET) < 0) {
```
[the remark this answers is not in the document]
atheriel: From what I can glean from `lseek(2)`, the only error condition applicable here is when `fd` is not an open file descriptor. Otherwise, I'm not sure.

src/webapplication.cpp
atheriel: Do we want to advertise range request support even if multipart ranges are not supported?

src/filedatasource.h
```diff
@@ -20,6 +20,7 @@ class FileDataSource : public DataSource {
 #else
   int _fd;
   off_t _length;
```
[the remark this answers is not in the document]
atheriel: Perhaps `_payloadSize`?

src/filedatasource-win.cpp
```diff
@@ -63,6 +63,10 @@ uint64_t FileDataSource::size() const {
   return _length.QuadPart;
 }
 
+bool FileDataSource::setRange(size_t start, size_t end) {
+  return false;
```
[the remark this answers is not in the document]
atheriel: I do not have a Windows test environment at the moment -- is that a dealbreaker?

src/filedatasource-unix.cpp
```diff
@@ -52,6 +52,18 @@ uint64_t FileDataSource::size() const {
   return _length;
 }
 
+bool FileDataSource::setRange(size_t start, size_t end) {
```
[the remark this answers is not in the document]
atheriel: Fixed both. The caller verified `end >= start` before, but now it's also checked here.

src/filedatasource-unix.cpp
```diff
@@ -35,7 +35,7 @@ FileDataSourceResult FileDataSource::initialize(const std::string& path, bool ow
       return FDS_ISDIR;
     }
 
-    _length = info.st_size;
+    _length = _fsize = info.st_size;
```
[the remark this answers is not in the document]
atheriel: Fixed.

src/filedatasource.h
```diff
@@ -32,6 +33,7 @@ class FileDataSource : public DataSource {
 
   FileDataSourceResult initialize(const std::string& path, bool owned);
   uint64_t size() const;
```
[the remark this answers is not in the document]
atheriel: I've added a new `fileSize()` method and documented both to clear up this ambiguity.

src/webapplication.cpp
```diff
@@ -563,6 +564,42 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
     }
   }
 
+  // Check the HTTP Range header, if it has been specified.
+  std::smatch rangeMatch;
+  int start = 0, end = pDataSource->size() - 1;
```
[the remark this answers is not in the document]
atheriel: Fixed.

src/webapplication.cpp
atheriel: I've added a specific `fileSize()` method and used it here to clear up the ambiguity.

src/webapplication.cpp
atheriel: Added a test for this :+1:.

src/webapplication.cpp
atheriel: I cleaned up the logic here so that we don't need to do this anymore.

src/webapplication.cpp
```diff
@@ -578,11 +615,22 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
...
+    hasRange = false;
   }
 
   if (client_cache_is_valid) {
     pDataSource2.reset();
     status_code = 304;
+    hasRange = false;
+  }
+
+  std::ostringstream contentRange;
```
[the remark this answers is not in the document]
atheriel: Fixed.

tests/testthat/test-static-paths.R
atheriel: It's not possible to ask for a 0-size range, but I've added tests for 1-byte reads from the start and end of the files (plus the existing ones that fail for last byte + 1).

tests/testthat/test-static-paths.R
atheriel: Done.

src/webapplication.cpp
```diff
@@ -1,6 +1,7 @@
 #include <boost/bind.hpp>
 #include <boost/shared_ptr.hpp>
 #include <boost/make_shared.hpp>
+#include <regex>
```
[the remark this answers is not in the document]
atheriel: I will attempt to convert it to something simpler using `find()`.

src/filedatasource-win.cpp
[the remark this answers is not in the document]
wch: Yes -- no Windows support for this would be a dealbreaker. Whenever possible, we don't want to have features that work on one platform and not another. We want code to work consistently across platforms -- it's a bad experience if someone's code works on one computer, but not another. For example, imagine if static file serving worked on Unixes but not Windows. Then someone would complain about it, and then someone is going to be on the hook to implement it.

If there's some technical reason that a feature can't work on Windows, that's one thing. For example, the Windows process model doesn't use `fork`, so some parallel processing code won't work. That's a legitimate reason for only supporting Unix -- but even so, it's an annoying one for users to have to work around (speaking from experience). If there's not a technical reason like that, then the code really should be made to work on all platforms.

src/webapplication.cpp
```diff
@@ -563,6 +564,42 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
...
 
+  // Check the HTTP Range header, if it has been specified.
+  std::smatch rangeMatch;
+  int start = 0, end = pDataSource->size() - 1;
+  if (pRequest->hasHeader("Range")) {
+    std::string rangeHeader = pRequest->getHeader("Range");
+    // This is a limited form of the HTTP Range header, so don't complain to the
+    // client if it's malformed -- just silently ignore it instead.
+    if (std::regex_search(rangeHeader, rangeMatch, std::regex("bytes=(\\d+)-(\\d+)?$"))) {
+      start = std::stoi(rangeMatch[1].str().c_str());
```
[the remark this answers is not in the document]
atheriel: I've updated parsing to use C's `strtoull()` (which doesn't throw) and added handling (and many tests) for malformed input.

src/webapplication.cpp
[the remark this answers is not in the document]
atheriel: I've removed `<regex>` use in favour of more hand-rolled parsing. Since `boost::regex` isn't bundled with the `BH` package I thought this was easier, if more verbose.

src/filedatasource-win.cpp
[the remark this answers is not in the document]
atheriel: That's a fair assessment, but it will take me some time. I'll try to resolve the other issues with this PR in the meantime.

src/webapplication.cpp
atheriel: Done.

src/webapplication.cpp
```diff
@@ -563,6 +564,42 @@ boost::shared_ptr<HttpResponse> RWebApplication::staticFileResponse(
     }
   }
 
+  // Check the HTTP Range header, if it has been specified.
```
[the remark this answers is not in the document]
atheriel: It's a bit awkward to do so, since there are currently three variables set in this section.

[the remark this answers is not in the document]
wch: It would be simpler to review and maintain the code if the range header parsing was a separate function. Right now the string parsing is mixed in with the logic for setting range and generating error responses; if the string parsing were done in a separate step, that would make the code more manageable. Even if the code is correct now, I would feel uneasy about touching it in the future.

Something like this would be easier to review and maintain:

```c++
std::pair<int64_t, int64_t> parseRangeHeader(std::string rangeHeader) {
  if (rangeHeader.find("bytes=") != 0) {
    return std::make_pair<int64_t, int64_t>(-1, -1);
  }

  rangeHeader = rangeHeader.substr(6);
  int sep_offset = rangeHeader.find("-");
  if (sep_offset == std::string::npos) {
    return std::make_pair<int64_t, int64_t>(-1, -1);
  }

  std::string start_str = rangeHeader.substr(0, sep_offset+1);
  std::string end_str   = rangeHeader.substr(sep_offset+1);
  
  int64_t start;
  int64_t end;

  // TODO: Do number parsing here; return (-1, -1) if either start or end are not
  // well-formed.
  
  return std::make_pair<int64_t, int64_t>(start, end);
}
```

The number parsing could be done with `strtoull()` as you currently have, or with `sscanf()`. (There are a number of details to get right with those functions and I don't have the time right now to try to do it, so I didn't implement that part.) It might make sense to extract it out into a function.

sounkou-bioinfo:
Hi all,
What is the status of this PR, this would be extremely useful in contexts like serving parquet files or other files for which partial downloads are supported. Use case do work on the client side instead of server for sat parquet files or serving duckdb databases

PR #297, opened 2021-03-24
Remove BH dependency

jcheng5:
Replaced boost/optional with https://github.com/akrzemi1/Optional.

`parse_http_date_string()` used boost/date_time; replaced this with `sscanf()` and `timegm()`. But `timegm()` isn't available on mingw, and the [first workaround I tried](https://stackoverflow.com/a/29411795) caused the DLL not to load during CI. The final workaround reimplements `timegm()` using Win32 APIs.

See also https://github.com/r-lib/later/pull/147


src/utils.h
```diff
@@ -303,23 +303,16 @@ inline time_t parse_http_date_string(const std::string& date) {
...
 
   // Format is now something like "21 10 2015 07:28:00"
-  // Convert to a boost::posix_time::ptime object
-  const static std::locale time_format = std::locale(
-    std::locale::classic(),
-    // Apparently it's not necessary to delete the time_input_facet.
-    new boost::posix_time::time_input_facet("%d %m %Y %H:%M:%S")
+  struct tm g = {0};
+  int res = sscanf(date_str.c_str(), "%d %d %d %d:%d:%d",
+    &g.tm_mday, &g.tm_mon, &g.tm_year, &g.tm_hour, &g.tm_min, &g.tm_sec
```
wch, line 308: This use of `sscanf` makes me nervous. See https://stackoverflow.com/questions/9457325/how-to-use-sscanf-correctly-and-safely

Maybe we can use `std::stoi`?

wch: Here's a rough sketch I tested out using cpp11, but obviously you'd just need to do this with C++. Note that it would need a try-catch, since this function can be called on the background thread.

```R
library(cpp11)
cpp_function('integers parse_date(std::string s) {
  writable::integers out(6);

  int pos = s.find(" ");
  out[0] = std::stoi(s.substr(0, pos));
  
  s = s.substr(pos + 1);
  pos = s.find(" ");
  out[1] = std::stoi(s.substr(0, pos));

  s = s.substr(pos + 1);
  pos = s.find(" ");
  out[2] = std::stoi(s.substr(0, pos));

  s = s.substr(pos + 1);
  pos = s.find(":");
  out[3] = std::stoi(s.substr(0, pos));
  
  s = s.substr(pos + 1);
  pos = s.find(":");
  out[4] = std::stoi(s.substr(0, pos));
  
  s = s.substr(pos + 1);
  out[5] = std::stoi(s);
  
  return out;
}')

parse_date("21 10 2015 07:28:00")
#> [1]   21   10 2015    7   28    0

parse_date("21 10 2015 07:28:a")
#> Error: stoi: no conversion
```

wch: I did a little research and I think this should work:

```C++
  std::istringstream date_ss(date_str);
  date_ss >> std::get_time(&t, "%d %m %Y %H:%M:%S");
  if (date_ss.fail()) {
    return 0;
  }
```

Reference: https://en.cppreference.com/w/cpp/io/manip/get_time

jcheng5: @wch Nice! Does this need a try/catch?

wch: I'm not entirely sure. It says [here](https://www.cplusplus.com/reference/iomanip/get_time/) that `get_time` can throw an exception, but I think something would have to be set up explicitly for that to occur:

> If the operation sets an internal state flag on the stream that was registered using its member exceptions, the function throws an exception of type ios_base::failure.

I found an example on SO where `get_time` throws an exception, but it seems like weird case:

> https://stackoverflow.com/questions/31030446/stdget-time-fails-at-midnight-or-noon-using-i-format-specifier

Just to be safe, I think we should use a try-catch.

jcheng5:
`get_time` does seem much, much better! Thanks for the tip.

For testing, I tried replacing the locale (`"C"`) with a locale that is not installed, and that caused an exception; the `try`/`catch(...)` catches that exception. I also tried replacing the locale with a valid locale in a different language, and that caused the parsing to fail gracefully (which confirms that `get_time` does respect the imbued locale).

wch:
To answer my own question: the day of week gets put in the `tm` object:

```R
cpp11::cpp_source(
  code = '
#include <cpp11.hpp>
#include <sstream>
#include <iomanip>

using namespace cpp11;

[[cpp11::register]]
cpp11::integers datetest(std::string date) {
  std::tm t = {};

  std::istringstream date_ss(date);
  date_ss >> std::get_time(&t, "%a, %d %b %Y %H:%M:%S GMT");
  return writable::integers({
    t.tm_isdst,
    t.tm_yday,
    t.tm_wday,
    t.tm_year,
    t.tm_mon,
    t.tm_mday,
    t.tm_hour,
    t.tm_min,
    t.tm_sec
  });
}
')

datetest("Tue, 30 Mar 2021 15:14:00 GMT")
#> [1]   0   0   2 121   2  30  15  14   0
datetest("Wed, 30 Mar 2021 15:14:00 GMT")
#> [1]   0   0   3 121   2  30  15  14   0
```

But I think it gets ignored for date-time comparisons later on.

DESCRIPTION
```diff
@@ -34,7 +35,7 @@ Imports:
     R6,
     promises,
     later (>= 0.8.0)
-LinkingTo: Rcpp, BH, later
+LinkingTo: Rcpp, later
 URL: https://github.com/rstudio/httpuv
 SystemRequirements: GNU make
```
wch, line 40:
```suggestion
SystemRequirements: GNU make, C++11
```

NEWS.md
```diff
@@ -1,3 +1,8 @@
+httpuv 1.5.5.9000
+=================
+
+* Remove BH dependency. (#297)
```
wch, line 4:
```suggestion
* Remove BH dependency. httpuv now requires a compiler which supports C++11. (#297)
```

wch:
As of gcc 4.9.3 (which I believe is the version included with rtools 3.5), `get_time` is not implemented. This is item 27.7 in the table here: https://gcc.gnu.org/onlinedocs/gcc-4.9.3/libstdc++/manual/manual/status.html

src/utils.cpp
```diff
@@ -40,3 +40,125 @@ std::string log_level(const std::string& level) {
...
+// @param input The istream to parse from
+// @param bytes The exact number of bytes to read from the input. If this many
+//   bytes are not available, false is returned.
+// @param Array where each element is a char* of length `bytes`; the final
+//   element must be NULL, or else the function will probably segfault.
+// @param pRes If true is returned, then this will be set to the index of the
+//   element in `values` that matched the input. If false is returned, then res
+//   will be untouched.
+// @return true if successful, false if reading failed or no match found
+bool str_read_lookup(std::istream* input, size_t bytes, const char** values, int* pRes) {
```
wch, line 82: No need to change this, but just a question. This could work with something like the following, right?

```C++
std::vector<std::string> MONTHS {"Jan", "Feb", "Mar", "Apr", "May", "Jun",
                                 "Jul", "Aug", "Sep", "Oct", "Nov", "Dec" };


bool str_read_lookup(std::istream* input, std::vector<std::string> values, int* pRes) {
  ...
}
```

wch (changes requested):
One piece of code that should be changed, but otherwise looks good.

src/utils.cpp
jcheng5: I didn't know about that, that's pretty nice!

PR #300, opened 2021-05-06
Add `timegm2()` function

wch:
BDR sent a note saying that `timegm()` is a non-standard GNU extension and it should be removed. This PR implements and uses `timegm2()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 httpuv 1.6.0.9000
 =================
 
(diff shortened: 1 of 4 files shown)


jcheng5:
Just curious if you considered this approach https://stackoverflow.com/a/58037981

wch:
Interesting, I didn't come across that. I'm a little surprised the code is as simple as it is.

It would be preferable to use a self-contained function like that, which doesn't call `setenv()` (for thread safety). I'll take a look at some alternatives.

PR #304, opened 2021-06-11
Don't send Content-Length header on 101 response

### LINKED ISSUE #303: Shiny is Returning Content-Length with a 101 (protocol switch)
While testing the use of YARP (https://github.com/microsoft/reverse-proxy) as a reverse proxy so that I could bundle a Shiny app as part of a larger ASP.NET Core web app I ran into an issue when the app attempted to connect its web socket.  The issue was related to Shiny returning a content-length of 0 alongside a 101.  Here is the issue I filed on YARP's GitHub: https://github.com/microsoft/reverse-proxy/issues/1073.  Repro steps are listed there as well.

According to the folks at Microsoft, Shiny is at fault because a response of 101 should not include a content-length header (https://github.com/microsoft/reverse-proxy/issues/1073#issuecomment-858977134).

### System details
Yarp.ReverseProxy 1.0.0-preview.11.21223.5
R version 4.0.3
Shiny pkg 1.6.0.
MacOS Mojave

jcheng5:
Fixes #303

### DIFF
--- a/src/httpresponse.cpp
+++ b/src/httpresponse.cpp
@@ -77,7 +77,7 @@ void HttpResponse::writeResponse() {
   // since they can't have a body.
   //
   // See: https://tools.ietf.org/html/rfc7230#section-3.3.2
-  if (_pBody != nullptr && !hasContentLengthHeader) {
+  if (_pBody != nullptr && !hasContentLengthHeader && _statusCode != 101) {
     response << "Content-Length: " << _pBody->size() << "\r\n";
   }
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@ httpuv 1.6.1.9000
 
 * Fixed #282: `startPipeServer()` failed with "invalid argument" error after update to libuv 1.37.0. (#283)
 
+* Fixed #303: "101 Switching Protocols" response should not have Content-Length header. (#304)
+
 httpuv 1.6.1
 ============
 


jcheng5:
(I just happened to be mucking around in this area of the code, on unrelated business)

src/httpresponse.cpp
```diff
@@ -77,7 +77,7 @@ void HttpResponse::writeResponse() {
   // since they can't have a body.
   //
   // See: https://tools.ietf.org/html/rfc7230#section-3.3.2
-  if (_pBody != nullptr && !hasContentLengthHeader) {
+  if (_pBody != nullptr && !hasContentLengthHeader && _statusCode != 101) {
```
wch, line 80: I think it would make more sense to fix the place where we create the 101 response object, so that the `_pBody` is a `nullptr`. Then we wouldn't have to special-case the 101 code here.


I believe that replacing `pDS` below with `std::shared_ptr<InMemoryDataSource>(nullptr)` would do the trick:

https://github.com/rstudio/httpuv/blob/13bcd461e4228583b29e63d4eae8ad6bb00a4ab0/src/httprequest.cpp#L796-L800

jcheng5: Unfortunately we do need to have a body, we use it to conduct the websocket handshake. Technically, it's not an HTTP body, but the first few bytes being spoken of a new conversation after the protocol switch has occurred--but it's convenient for us to reuse the body for this purpose.

But maybe a flag on the response object indicating that the body object should be considered not to be a "real" HTTP body? Either way, I think it'll be special case code somewhere, since nothing else in HTTP/1.1 acts like this AFAIK.

jcheng5:
This PR will actually conflict with #305, and #305 contains its own fix for this issue. So if you're seriously considering the gzip PR, let's close this one.

PR #305, opened 2021-06-12
Implement gzip compression

### LINKED ISSUE #303: Shiny is Returning Content-Length with a 101 (protocol switch)

jcheng5:
(Incidentally, fixes #303)

A Connect customer was seeing significantly slower loading times (accessing through a VPN) than with a previous setup using Shiny Server. Shiny Server does its own gzip compression, but Connect just passes through whatever it gets from Shiny. It felt to me like httpuv should just do this.

This implementation uses zlib in streaming mode, so Content-Length headers are no longer returned (when gzipped); instead the response is sent using chunked encoding.

One open question with this PR is that it currently makes the decision about gzip vs. no gzip pretty late, like right before the headers are sent. One of the things it takes into consideration is whether there is a body. This means that HEAD requests for a resource that would be gzipped, returns different headers than the equivalent GET (the former returns `Content-Length`, the latter does not; and the latter returns `Content-Encoding: gzip` and `Transfer-Encoding: chunked` which the former does not). It'd be ideal if the headers were exactly the same.

### DIFF
--- a/src/Makevars.win
+++ b/src/Makevars.win
@@ -2,7 +2,7 @@
 CXX_STD=CXX11
 
 PKG_LIBS = ./libuv/libuv.a ./http-parser/http_parser.o ./sha1/sha1.o ./base64/base64.o \
-	-lpthread -lws2_32 -lkernel32 -lpsapi -liphlpapi -lshell32 -luserenv
+	-lpthread -lws2_32 -lkernel32 -lpsapi -liphlpapi -lshell32 -luserenv -lz
 
 PKG_CFLAGS = $(C_VISIBILITY) -DSTRICT_R_HEADERS
 PKG_CXXFLAGS = $(CXX_VISIBILITY) -DSTRICT_R_HEADERS
--- a/src/gzipdatasource.cpp
+++ b/src/gzipdatasource.cpp
@@ -1,4 +1,5 @@
 #include "gzipdatasource.h"
+#include "utils.h"
 
 GZipDataSource::GZipDataSource(std::shared_ptr<DataSource> pData) :
   _pData(pData), _state(Streaming) {
@@ -21,7 +22,7 @@ GZipDataSource::~GZipDataSource() {
 }
 
 uint64_t GZipDataSource::size() const {
-  std::cerr << "GZipDataSource::size() was called, this should never happen\n";
+  debug_log("GZipDataSource::size() was called, this should never happen\n", LOG_WARN);
   return 0;
 }
 
--- a/src/gzipdatasource.cpp
+++ b/src/gzipdatasource.cpp
@@ -5,6 +5,7 @@ GZipDataSource::GZipDataSource(std::shared_ptr<DataSource> pData) :
   _pData(pData), _state(Streaming) {
 
   _zstrm = {0};
+  _inputBuf = {0};
   int res = deflateInit2(&_zstrm, 4, Z_DEFLATED, 15 + 16, 8, Z_DEFAULT_STRATEGY);
   if (res != Z_OK) {
     if (_zstrm.msg) {
--- a/src/gzipdatasource.cpp
+++ b/src/gzipdatasource.cpp
@@ -1,0 +1,96 @@
+#include "gzipdatasource.h"
+
+GZipDataSource::GZipDataSource(std::shared_ptr<DataSource> pData) :
+  _pData(pData), _state(Streaming) {
+
+  _zstrm = {0};
+  int res = deflateInit2(&_zstrm, 4, Z_DEFLATED, 15 + 16, 8, Z_DEFAULT_STRATEGY);
+  if (res != Z_OK) {
+    if (_zstrm.msg) {
+      throw std::runtime_error(_zstrm.msg);
+    } else {
+      throw std::runtime_error("zlib initialization failed");
+    }
+  }
+}
+
+GZipDataSource::~GZipDataSource() {
+  freeInputBuffer(true);
+  // ignore errors on destruction
+  deflateEnd(&_zstrm);
+}
+
+uint64_t GZipDataSource::size() const {
+  std::cerr << "GZipDataSource::size() was called, this should never happen\n";
+  return 0;
+}
+
+uv_buf_t GZipDataSource::getData(size_t bytesDesired) {
+  if (_state == Done) {
+    // GZip stream written, nothing more to do
+    return {0};
+  }
+
+  // Prepare the output area to be written to
+  Bytef* outputBuf = (Bytef*)malloc(bytesDesired);
+  _zstrm.next_out = outputBuf;
+  _zstrm.avail_out = bytesDesired;
+
+  // There's room to write, and things we need to write: if Streaming, then
+  // there's potentially more data; and if Finishing, then we need to write
+  // the gzip footer.
+  while (_zstrm.avail_out > 0 && _state != Done) {
+    if (_state == Streaming && _zstrm.avail_in == 0) {
+      freeInputBuffer();
+
+      _inputBuf = _pData->getData(bytesDesired);
+      _zstrm.next_in = (Bytef*)_inputBuf.base;
+      _zstrm.avail_in = _inputBuf.len;
+
+      if (_inputBuf.len == 0) {
+        _state = Finishing;
+      }
+    }
+
+    deflateNext();
+  }
+
+  freeInputBuffer();
+
+  uv_buf_t ret = {0};
+  ret.base = (char*)outputBuf;
+  ret.len = bytesDesired - _zstrm.avail_out;
+  return ret;
+}
+
+void GZipDataSource::freeData(uv_buf_t buffer) {
+  free(buffer.base);
+}
+
+void GZipDataSource::close() {
+  _pData->close();
+}
+
+// Attempt to deflate more data, reading from _zstrm.next_in and writing to
+// _zstrm.next_out. Both reads and (potentially) writes _state.
+void GZipDataSource::deflateNext() {
+  int res = deflate(&_zstrm, (_state == Finishing) ? Z_FINISH : Z_NO_FLUSH);
+  if (res == Z_STREAM_END) {
+    _state = Done;
+  } else if (res != Z_OK) {
+    throw std::runtime_error("deflate failed!");
+  }
+}
+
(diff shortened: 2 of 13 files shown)


src/Makevars.win
```diff
@@ -2,7 +2,7 @@
 CXX_STD=CXX11
 
 PKG_LIBS = ./libuv/libuv.a ./http-parser/http_parser.o ./sha1/sha1.o ./base64/base64.o \
-	-lpthread -lws2_32 -lkernel32 -lpsapi -liphlpapi -lshell32 -luserenv
+	-lpthread -lws2_32 -lkernel32 -lpsapi -liphlpapi -lshell32 -luserenv -lz
```
jcheng5, line 5: I'm not sure why this isn't necessary in `Makevars` as well, but, it compiles fine without it...

src/gzipdatasource.cpp
```diff
@@ -0,0 +1,98 @@
+#include "gzipdatasource.h"
+#include "utils.h"
+
+GZipDataSource::GZipDataSource(std::shared_ptr<DataSource> pData) :
+  _pData(pData), _state(Streaming) {
+
+  _zstrm = {0};
+  _inputBuf = {0};
+  int res = deflateInit2(&_zstrm, 4, Z_DEFLATED, 15 + 16, 8, Z_DEFAULT_STRATEGY);
```
jcheng5, line 9:
```suggestion
  int res = deflateInit2(&_zstrm, 6, Z_DEFLATED, 15 + 16, 8, Z_DEFAULT_STRATEGY);
```

src/Makevars.win
wch: Probably because it's included in the system Makeconf file. On my Mac's /Library/Frameworks/R.framework/Versions/4.0/Resources/etc/Makeconf file, I see:

```
LIBS = -L/usr/local/lib -lpcre2-8 -llzma -lbz2 -lz -licucore -ldl -lm -liconv
```

Hopefully it's there on Solaris as well. 😬

src/httpresponse.cpp
```diff
@@ -62,23 +64,66 @@ void HttpResponse::writeResponse() {
...
+    // The response already has a Content-Encoding
+    gzip = false;
+  } else if (_statusCode == 101 || _pBody == nullptr) {
+    gzip = false;
+  } else {
+    RequestHeaders h = _pRequest->headers();
+    auto acceptEncoding = h.find("Accept-Encoding");
+    if (acceptEncoding != h.end()) {
+      std::string enc = acceptEncoding->second;
+      if (enc.find("gzip") != std::string::npos) {
```
wch, line 94: Do we also want to accept `"*"`?

I'm a little confused by the documentation, though. On [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) it says:
> Matches any content encoding not already listed in the header. This is the default value if the header is not present. It doesn't mean that any algorithm is supported; merely that no preference is expressed.

I don't understand that all.

In [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.3), it says this, which seems to make more sense:
> The special "*" symbol in an Accept-Encoding field matches any available content-coding not explicitly listed in the header field.

wch: Maybe it doesn't matter in practice: I checked with Safari, Firefox, and Chrome, and they all send this in their requests:

```R
Accept-Encoding: gzip, deflate, br
```

src/httpresponse.cpp
```diff
@@ -62,23 +64,66 @@ void HttpResponse::writeResponse() {
...
   }
 
-  // Some valid responses (such as HTTP 204 and 304) must not set this header,
-  // since they can't have a body.
-  //
-  // See: https://tools.ietf.org/html/rfc7230#section-3.3.2
-  if (_pBody != nullptr && !hasContentLengthHeader) {
+  if (gzip) {
+    response << "Content-Encoding: gzip\r\n";
+    _chunked = true;
```
wch, line 108: Just curious: why are we always using `Transfer-Encoding: chunked` when gzip is used?

src/uvutil.cpp
```diff
@@ -93,15 +130,41 @@ void ExtendedWrite::next() {
...
+  if (this->_chunked) {
+    if (buf.len == 0) {
+      // In chunked mode, the last chunk must be followed by one more "\r\n".
+      pWriteOp->setSuffix(TRAILER.c_str(), TRAILER.length());
+    } else {
+      // In chunked mode, data chunks must be preceded by 1) the number of bytes
+      // in the chunk, as a hexadecimal string; and 2) "\r\n"; and succeeded by
+      // another "\r\n"
+      char prefix[16];
+      memset(prefix, 0, sizeof(prefix));
```
wch, line 149: I _think_ this should work, and seems safer and cleaner than using `memset()`.

```suggestion
      char prefix[16] = {0};
```

src/uvutil.cpp
```diff
@@ -93,15 +130,41 @@ void ExtendedWrite::next() {
...
+      int len = snprintf(prefix, sizeof(prefix), "%lX\r\n", buf.len);
```
wch, line 150: Now that we're using C++11, you could use `to_string()`. Then the other prefix/suffix code would have to be updated to use `std::string`, which would make it a bit safer.

src/httpresponse.cpp
jcheng5: If we don’t use chunked, I believe we must provide Content-Length, and there’s no way to know that in advance, since the gzip compression is being done in a streaming way (rather than compressing the whole response body before we send the first byte).

src/uvutil.cpp
```diff
@@ -4,17 +4,44 @@
...
+
+  void setSuffix(const char* data, size_t len) {
+    suffix.clear();
+    std::copy(data, data + len, std::back_inserter(suffix));
+  }
+
+  std::vector<uv_buf_t> bufs() {
+    std::vector<uv_buf_t> res;
+    if (prefix.size() > 0) {
+      res.push_back(uv_buf_init(&prefix[0], prefix.size()));
```
wch, line 36: I want to make sure I understand: after this `uv_buf_init` is called, and as long as object returned by `bufs()` exists, it's not safe to modify `prefix`, correct? Because changing `prefix` could cause reallocation, and the `uv_buf_t` would then be pointing to unallocated memory. If that's right, can you add a comment about that to the top of this function?

src/uvutil.cpp
```diff
@@ -4,17 +4,44 @@
...
+public:
+  uv_write_t handle;
 
   WriteOp(ExtendedWrite* parent, uv_buf_t data)
         : pParent(parent), buffer(data) {
     memset(&handle, 0, sizeof(uv_write_t));
     handle.data = this;
   }
 
+  void setPrefix(const char* data, size_t len) {
```
wch, line 23: Just to be extra safe:

```suggestion
  void setPrefix(const char* data, size_t len) {
    ASSERT_BACKGROUND_THREAD()
```

src/uvutil.cpp
```diff
@@ -4,17 +4,44 @@
...
+    prefix.clear();
+    std::copy(data, data + len, std::back_inserter(prefix));
+  }
+
+  void setSuffix(const char* data, size_t len) {
```
wch, line 28:
```suggestion
  void setSuffix(const char* data, size_t len) {
    ASSERT_BACKGROUND_THREAD()
```

src/uvutil.cpp
```diff
@@ -4,17 +4,44 @@
...
+    suffix.clear();
+    std::copy(data, data + len, std::back_inserter(suffix));
+  }
+
+  std::vector<uv_buf_t> bufs() {
```
wch, line 33:
```suggestion
  std::vector<uv_buf_t> bufs() {
    ASSERT_BACKGROUND_THREAD()
```

src/uvutil.cpp
jcheng5: The number needs to be hex, which I don't see a way to do with `to_string()`. Would you prefer `std::stringstream`?

wch: Oh, that's too bad. I'm fine with keeping it as is instead of switching to `stringstream`.

src/Makevars.win
jcheng5: Here's what I see in R-4.1.0-solaris-ods.tar.gz from https://files.r-hub.io/solaris/, under lib/R/etc/Makeconf:

```
LIBS = -L/opt/csw/lib -lpcre2-8 -llzma -lbz2 -lz -lnsl -lsocket -lrt -lsunmath -ldl -lm -liconv -licuuc -licui18n
```

src/httpresponse.cpp
jcheng5: Yeah, I'm playing quite fast and loose here. Not respecting `gzip;q=0`, which should explicitly disallow gzip. Frankly, I just didn't want to go through the effort of fully parsing this header using C++.

src/uvutil.cpp
jcheng5: Hmmm, you're correct but it's only a problem because I'm setting the prefix via an accessor function instead of just taking the prefix in the constructor. Let me do that instead.

jcheng5: Alright yeah, it's much cleaner now. Thanks.

wch (approved):

wch:
Looks good! Can you add a NEWS entry and version bump?

jcheng5:
Oh, I forgot to address the issue I brought up in the last paragraph of the PR description, about HEAD and GET returning different headers. Maybe we can talk about this at our next 1:1.

PR #319, opened 2021-11-04
Fix builds when a newer automake is present

### LINKED ISSUE #280: Error Installing httpuv on docker alpine 3.12
I get an error when trying to install httpuv to my docker container based on an alpine 3.12 image
```
make --directory=libuv \
        HAVE_DTRACE=0
make[1]: Entering directory '/tmp/RtmpomCKmc/R.INSTALLe273c6e03ca/httpuv/src/libuv'
 cd . && /bin/sh /tmp/RtmpomCKmc/R.INSTALLe273c6e03ca/httpuv/src/libuv/missing automake-1.16 --foreign
configure.ac:21: error: version mismatch.  This is Automake 1.16.2,
configure.ac:21: but the definition used by this AM_INIT_AUTOMAKE
configure.ac:21: comes from Automake 1.16.1.  You should recreate
configure.ac:21: aclocal.m4 with aclocal and run automake again.
WARNING: 'automake-1.16' is probably too old.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <https://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <https://www.gnu.org/software/autoconf>
         <https://www.gnu.org/software/m4/>
         <https://www.perl.org/>
make[1]: *** [Makefile:1471: Makefile.in] Error 1
make[1]: Leaving directory '/tmp/RtmpomCKmc/R.INSTALLe273c6e03ca/httpuv/src/libuv'
make: *** [Makevars:73: libuv/.libs/libuv.a] Error 2
ERROR: compilation failed for package ‘httpuv’
* removing ‘/usr/lib/R/library/httpuv’

The downloaded source packages are in
        ‘/tmp/RtmpmFIjab/downloaded_packages’
Updating HTML index of packages in '.Library'
Making 'packages.html' ... done
Warning message:
In install.packages("httpuv") :
  installation of package ‘httpuv’ had non-zero exit status

```

This automake issue seems to have been identified before [here](https://github.com/rstudio/httpuv/issues/181) and [here](https://github.com/rstudio/httpuv/issues/63). I can't however find a work around for this.

wch:
This is fixes #280.

Tested on Mac, and on an Alpine container build with this Dockerfile:

```Dockerfile
FROM rhub/r-minimal:latest

RUN apk update && \
    apk add libcurl && \
    apk add g++ && \
    apk add gcc && \
    apk add linux-headers && \
    apk add gfortran && \
    apk add automake libtool m4 autoconf linux-headers && \
    apk add openssl

RUN apk add --no-cache \
    autoconf=2.69-r2 \
    automake=1.16.2-r0 \
    build-base=0.5-r2 \
    R \
    R-dev && \
    apk add --no-cache --virtual .build-deps \
    wget=1.20.3-r1 && \
    apk del .build-deps

RUN echo 'options(\n\
  repos = c(CRAN = "https://cloud.r-project.org/"),\n\
  download.file.method = "libcurl",\n\
  # Detect number of physical cores\n\
  Ncpus = parallel::detectCores(logical=FALSE)\n\
)' >> /etc/R/Rprofile.site

RUN R -e "install.packages('remotes', repos = c(CRAN = 'https://cloud.r-project.org')); \
    install.packages('renv', repos = c(CRAN = 'https://cloud.r-project.org')); \
    renv::restore(repos = c(CRAN = 'https://cloud.r-project.org')); \
    renv::clean();"

RUN R -e "install.packages(c('Rcpp', 'R6', 'promises', 'later'), repos = c(CRAN = 'https://cloud.r-project.org') )"

CMD ["/bin/bash"]
```

To build the Docker image, run it, and install httpuv:

```
docker build -t alpinehttpuv .
docker run --rm -ti alpinehttpuv
R
remotes::install_github('rstudio/httpuv@fix-automake')
```

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -76,6 +76,8 @@ libuv/Makefile: libuv/m4/lt~obsolete.m4
 		echo "automake not found. Touching files so configure will not try to run automake."; \
 		touch aclocal.m4; \
 		touch -r aclocal.m4 configure Makefile.in; \
+	else \
+		touch configure.ac; \
 	fi; \
 	chmod +x configure; \
 	CC="$(CC)" CFLAGS="$(CFLAGS) $(CPICFLAGS) $(C_VISIBILITY)" AR="$(AR)" RANLIB="$(RANLIB)" LDFLAGS="$(LDFLAGS)" ./configure $(CONFIGURE_FLAGS)
--- a/src/Makevars
+++ b/src/Makevars
@@ -50,10 +50,18 @@ $(SHLIB): libuv/.libs/libuv.a http-parser/http_parser.o sha1/sha1.o base64/base6
 libuv/m4/lt~obsolete.m4: libuv/m4/lt_obsolete.m4
 	cp -p -f libuv/m4/lt_obsolete.m4 libuv/m4/lt~obsolete.m4
 
-# Run ./configure. We need to touch various autotools-related files to avoid
-# it trying to run autotools programs again. We also need to make sure
-# configure is executable, because on some platforms, calling unzip() in R
-# does not preserve the executable bit.
+# Run ./configure to create the Makefile.
+#
+# On systems that do _not_ have automake installed, we need to make sure that
+# configure does not try to run automake, because it will fail. To do that, we
+# we need to touch various autotools-related files so it doesn't try to run
+# autotools programs again. We also need to make sure configure is executable,
+# because on some platforms, calling unzip() in R does not preserve the
+# executable bit.
+#
+# If the system does have automake, then we'll just run configure; it may in
+# turn run automake if the version of automake on the system is newer than the
+# one used when the configure file was created.
 #
 # It's VERY IMPORTANT that mtime(aclocal.m4) <= mtime(configure), and also
 # mtime(aclocal.m4) <= mtime(Makefile.in). On some platforms, passing multiple
--- a/src/Makevars
+++ b/src/Makevars
@@ -89,8 +89,8 @@ libuv/Makefile: libuv/m4/lt~obsolete.m4
 		touch aclocal.m4; \
 		touch -r aclocal.m4 configure Makefile.in; \
 	else \
-		echo "automake found. Touching configure.ac."; \
-		touch configure.ac; \
+		echo "automake found. Running autogen.sh."; \
+		sh autogen.sh; \
 	fi; \
 	chmod +x configure; \
 	CC="$(CC)" CFLAGS="$(CFLAGS) $(CPICFLAGS) $(C_VISIBILITY)" AR="$(AR)" RANLIB="$(RANLIB)" LDFLAGS="$(LDFLAGS)" ./configure $(CONFIGURE_FLAGS)
--- a/src/Makevars
+++ b/src/Makevars
@@ -59,9 +59,9 @@ libuv/m4/lt~obsolete.m4: libuv/m4/lt_obsolete.m4
 # because on some platforms, calling unzip() in R does not preserve the
 # executable bit.
 #
-# If the system does have automake, then we'll just run configure; it may in
-# turn run automake if the version of automake on the system is newer than the
-# one used when the configure file was created.
+# If the system does have automake, then we'll run autogen.sh before configure,
+# as per the official build instructions for libuv. autogen.sh will in turn run
+# aclocal, autoconf, and automake.
 #
 # It's VERY IMPORTANT that mtime(aclocal.m4) <= mtime(configure), and also
 # mtime(aclocal.m4) <= mtime(Makefile.in). On some platforms, passing multiple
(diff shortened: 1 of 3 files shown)


src/Makevars
```diff
@@ -63,11 +63,14 @@ libuv/m4/lt~obsolete.m4: libuv/m4/lt_obsolete.m4
 # use "-r aclocal.m4" to ensure that all three files are guaranteed to have
 # precisely the same timestamp value.
 libuv/Makefile: libuv/m4/lt~obsolete.m4
-	(cd libuv \
-		&& touch aclocal.m4 \
-		&& touch -r aclocal.m4 configure Makefile.in \
-		&& chmod +x configure \
-		&& CC="$(CC)" CFLAGS="$(CFLAGS) $(CPICFLAGS) $(C_VISIBILITY)" AR="$(AR)" RANLIB="$(RANLIB)" LDFLAGS="$(LDFLAGS)" ./configure $(CONFIGURE_FLAGS))
+	cd libuv; \
```
jcheng5, line 66: This makes `cd` “permanent” whereas the parens kept the effects of the `cd` inside the parens, is this safe?

wch: I initially thought that would be an issue as well, but it turns out it's not. In a Makefile, each line (which isn't a continuation of the previous line with `\`) is run in a new shell.

Here's an example Makefile which illustrates:

```Makefile
foo:
	cd src; pwd
	pwd
	cd src; pwd
```

If you run it:

```
$ make foo
cd src; pwd
/Users/winston/Projects/httpuv/src
pwd
/Users/winston/Projects/httpuv
cd src; pwd
/Users/winston/Projects/httpuv/src
```

This is also the reason that all the lines are continued with a `\`. If they weren't, then the `if` statement wouldn't work at all (since each line would be run in a new shell), and each line would have to start with `cd libuv; ....`;

jcheng5: TIL

jcheng5:
Ok so, this now guarantees that automake will not try to run if it’s not installed. If automake is installed, isn’t it kind of nondeterministic whether automake runs? Should we delete Makefile in that case?

wch:
What we've observed is:
* If `automake` is not installed, it will not run.
* If `automake` is installed and is a greater version than used to create the `configure` script, it will run.
* If `automake` is installed and the same version as was used to create the `configure` script, it won't run.


I'm not sure what will happen if `automake` is an older version than was used to create the `configure` script. I suspect it won't run, but I'm not 100% sure. 

Is this the nondeterminism you're referring to, or is there something else?

jcheng5:
Nondeterministic depending on the timestamps of aclocal.m4, configure, and Makefile.in; if automake is installed and the same version, I would expect it to run sometimes. I think that's the reason we had to put that `touch` command there in the first place; sometimes it worked, sometimes it didn't. Or am I misremembering?

wch:
Interesting... So perhaps we should have an `elif` condition where it does `touch aclocal.m4`, and nothing else? I think this should cause `automake` to run consistently (if it's a newer version).

jcheng5:
@wch From the diagrams [here](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Making-configure-Scripts.html) it looks like it should be `touch configure.ac`?

wch:
For future reference, this is the current state of the `Makevars`:

```Makefile
	if ! command -v automake >/dev/null 2>&1 ; then \
		echo "automake not found. Touching files so configure will not try to run automake."; \
		touch aclocal.m4; \
		touch -r aclocal.m4 configure Makefile.in; \
	else \
		echo "automake found. Touching configure.ac."; \
		touch configure.ac; \
	fi; \
```

This works on:
* My Mac, which doesn't have `automake` or `autoconf`
* The Alpine Docker image, which has a newer version of `automake` and has `autoconf`
* The `wch1/r-devel` Docker image, which has `automake` and `autoconf` 
* Ubuntu 20.04 in GHA (I don't know the status of `automake` and `autoconf`)

It fails on:
* Ubuntu 18.04 in GHA (I don't know the status of `automake` and `autoconf`)

The message is:

```
  OE> CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/bash /tmp/RtmpH9ZvQr/R.INSTALL22a81f3b8dcf/httpuv/src/libuv/missing aclocal-1.16 -I m4
  OE> /tmp/RtmpH9ZvQr/R.INSTALL22a81f3b8dcf/httpuv/src/libuv/missing: line 81: aclocal-1.16: command not found
  OE> WARNING: 'aclocal-1.16' is missing on your system.
  OE>          You should only need it if you modified 'acinclude.m4' or
  OE>          'configure.ac' or m4 files included by 'configure.ac'.
  OE>          The 'aclocal' program is part of the GNU Automake package:
  OE>          <gnu.org/software/automake>
  OE>          It also requires GNU Autoconf, GNU m4 and Perl in order to run:
  OE>          <gnu.org/software/autoconf>
  OE>          <gnu.org/software/m4>
  OE>          <perl.org>
  OE> Makefile:1499: recipe for target 'aclocal.m4' failed
```

I suspect that the reason it's failing on Ubuntu 18.04 on GHA is because it has `automake` but not `autoconf`, but I'll have to do some work to check this.

wch:
OK, I think I've figured it out. The key is that if `automake` is installed, you just want to follow the official libuv build instructions, which is to run `sh autogen.sh` before `./configure`.

The `autogen.sh` script runs `aclocal`, `autoconf`, and `automake`, and in doing so, it finds the right version of `aclocal`.

wch:
This is passing on all platforms in GHA, and it also passes locally with:
* My Mac (no `automake`)
* The Alpine Docker image
* The `wch1/r-devel` Docker image

So I think we're good, finally.

jcheng5 (approved):

PR #324, opened 2021-12-14
Add support for R on Windows UCRT

wch:
I received the following email from Tomas Kalibera about supporting R on Windows UCRT.

Note: So far this PR simply applies the patch provided by CRAN, but I think it will cause problems for non-UCRT R on Windows.

*****

Dear maintainers,

This concerns the CRAN packages

   Cairo cepreader gpboost httpuv ipaddress lightgbm proj4 prophet
   RcppCWB RcppParallel RDieHarder re2 redux rgeolocate RGtk2 tth
   udunits2 unrtf

maintained by one of you:
....

your packages need to be updated for R-devel/R 4.2 to work on Windows,
following the recent switch to UCRT and Rtools42.

Sorry for the group message, please feel free to respond individually
regarding your package or ask specifically about what needs to be fixed.

I've created patches for you, so please review them and fix your packages:

https://svn.r-project.org/R-dev-web/trunk/WindowsBuilds/winutf8/ucrt3/r_packages/patches/CRAN/

You can apply them as follows

tar xfz package_1.0.0.tar.gz

wget
https://svn.r-project.org/R-dev-web/trunk/WindowsBuilds/winutf8/ucrt3/r_packages/patches/CRAN/package.diff

patch --binary < package.diff

These patches are currently automatically applied by R-devel on Windows
at installation time, which makes most of your packages pass their
checks (as OK or NOTE), but please check your results carefully and
carefully review the patches. Usually these changes were because of
newer GCC or newer MinGW in the toolchain, but some for other reasons,
and some of them will definitely have to be improved so that the package
keeps building also for older versions of R using Rtools40. We have only
been testing the patches with UCRT (and Rtools42) on Windows.

For more information, please see

https://developer.r-project.org/Blog/public/2021/12/07/upcoming-changes-in-r-4.2-on-windows/
https://developer.r-project.org/WindowsBuilds/winutf8/ucrt3/howto.html

Once you add your patches/fix the issues, your package will probably
show a warning during R CMD check (as patching would be attempted to be
applied again). That's ok, at that point please let me know and I will
remove my patch from the repository of automatically applied patches.

If you end up just applying the patch as is, there is probably no need
testing on your end, but you can do so using Winbuilder, r-hub, github
actions (e.g. https://github.com/kalibera/ucrt3).

If you wanted to test locally on your Windows machine and do not have a
UCRT version of R-devel yet, please uninstall your old version of
R-devel, delete the old library used with that, install a new UCRT
version of R-devel , and install Rtools42. You can keep Rtools40
installed if you need it with R 4.1 or earlier.

Currently, the new R-devel can be downloaded from
https://www.r-project.org/nosvn/winutf8/ucrt3/web/rdevel.html

And Rtools42 from
https://www.r-project.org/nosvn/winutf8/ucrt3/web/rtools.html

If you end up testing locally, you can use _R_INSTALL_TIME_PATCHES_
environment variable to disable the automated patching, see the "howto"
document above. That way you could also see what the original issue was
causing.

If you wanted to find libraries to link for yourself, e.g. in a newer
version of your package, please look for "Using findLinkingOrder with
Rtools42 (tiff package example)" in the "howto" document above. I
created the patches for you manually before we finished this script, so
you may be able to create a shorter version using it, but - it's
probably not worth the effort.

If you wanted to try in a virtual machine, but did not have a license,
you can use also an automated  setup of a free trial VM from
https://developer.r-project.org/Blog/public/2021/03/18/virtual-windows-machine-for-checking-r-packages

(but that needs a very good and un-metered network connection to install)

Please let us know if you have any questions.

Thanks,
Tomas & Uwe


wch:
Merging this now, since it doesn't seem to make anything worse on existing platforms.

PR #326, opened 2021-12-16
Test on windows ucrt

schloerke:


wch:
For platforms other than Windows-UCRT: I just realized that I don’t know if the `-I libuv/include/` puts the local copy first on the search path, or last. If it’s last, that would be problematic, because then it would compile with the system libuv headers (if they’re present), but then statically link to the local `libuv.a`. We need to make sure that the local path is searched first.

wch:
Regarding the include dir search path (https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html):

> There are a number of command-line options you can use to add additional directories to the search path. The most commonly-used option is `-Idir`, which causes `dir` to be searched after the current directory (for the quote form of the directive) and ahead of the standard system directories. You can specify multiple `-I` options on the command line, in which case the directories are searched in left-to-right order.

Since the Makevars and Makevars.win files have `-Ilibuv/include`, this means that the local headers will take precedence ahead of the system headers, which is good.

wch:
For future reference: the tests are failing on windows-devel because that build of R was built with UCRT, and so it uses `Makevars.ucrt` to build. The `Makevars.ucrt` file expects libuv to be available at the system level, which _would_ be true if we installed rtools42, but the GHA action is currently installing rtools40.

Using rtools42 is a bit difficult because it is very large (400MB compressed, and a few GB uncompressed). For now we won't use rtools42 in GHA, and hopefully we'll have a better way of dealing with this in the future.

schloerke:
@nealrichardson The `libuv` header changes were merged in https://github.com/rstudio/httpuv/pull/327 (`httpuv` >= v1.6.5)

However, the GHA UCRT changes have been dropped.

-----------------------

I'm going to update the branch to get the latest changes.

I pushed a change in https://github.com/rstudio/shiny-workflows/commit/7f01ec45324ff9b185a2b119e5a1715e163da4d8 to have windows devel test on R Tools 43 for R version 4.3.

`r-lib/actions` asks https://api.r-hub.io/rversions/resolve/release/linux-ubuntu-20.04 for the R version. Given https://github.com/r-lib/actions/blob/467c2b35913addb015a48dcbf1893017e6167e67/setup-r/src/installer.ts#L443-L449 , I believe the rtools versions are being installed. 

I see https://github.com/r-lib/actions/blob/467c2b35913addb015a48dcbf1893017e6167e67/setup-r/src/installer.ts#L514, but I don't know if it is reached given we are supplying the rtools version in `rstudio/shiny-workflows`.

-----------------------------------

@neal, I'm happy to hand it off to you here to take over the finish.

schloerke:
@nealrichardson Merging in the commits and with the latest `rstudio/shiny-workflows` commit that [runs windows on rtools 43, 42, and 40](https://github.com/rstudio/shiny-workflows/commit/7f01ec45324ff9b185a2b119e5a1715e163da4d8#diff-9c940e8ad2b7bc4c26ec3da57b94bc00e73e2166cfed689da51a4c59bcc0a310R143-R147), this PR is now empty.

Closing PR

PR #348, opened 2022-12-13
Fixes for older macOS

barracuda156:
This PR is based on patches used in Macports for building `libuv`: https://github.com/macports/macports-ports/tree/master/devel/libuv/files

P. S. I am bringing R packages into Macports now, and `httpuv` among them, and we want everything to build on Leopard up; it is desirable to have it fixed with upstream rather than keep carrying our local patch.

### DIFF
--- a/src/libuv/src/unix/core.c
+++ b/src/libuv/src/unix/core.c
@@ -531,18 +531,30 @@ int uv__accept(int sockfd) {
  * will unwind the thread when it's in the cancel state. Work around that
  * by making the system call directly. Musl libc is unaffected.
  */
+#if defined(__GNUC__)
+# define GCC_VERSION \
+	(__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
+#endif
+#if defined(__clang__) || (defined(GCC_VERSION) && (GCC_VERSION >= 40500))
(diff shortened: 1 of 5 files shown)


wch:
What versions of MacOS does this provide support for? The highest version I see reference to in this PR is `MAC_OS_X_VERSION_MAX_ALLOWED >= 1070`, and I believe that's for OS X 10.7, which last had a release 10 years ago.

It would be preferable if you could get the changes into libuv. We periodically update the copy of libuv bundled with httpuv, and it wouldn't be a good use of our time to be maintaining libuv patches for these old OS versions.

barracuda156:
@wch As for OS X versions, everything from 10.4.11 up is still used, and specifically 10.5.8 is used quite a bit, since that is the last official release supporting PPC.
However I agree, that if you just borrow the code from `libuv`, then you face the same problem as we do with carrying patches – and it won’t be a Pareto improvement :)

I will try to find out what happened with submission of those to `libuv` upstream. If no one tried for w/e reason, I can try that.

barracuda156:
@ryandesign Ryan, may I ask if you know whether our patches to `libuv` were submitted to upstream? Or should I try that?
At least two `R` packages are broken without those, and understandably it is preferable to have `libuv` itself fixed rather than carrying patches, if that works.

PR #352, opened 2023-01-11
Fix race condition between `Barrier` destructor and `Barrier.wait()`

### LINKED ISSUE #351: Random crashes on MacOS
The `curl` package has a function `curl_echo` which starts a temporary httpuv server. As of somewhat recently (several weeks/months), the examples and tests for this function are spuriously crashing on MacOS. The crashes appear on all versions of MacOS currently in GHA: 10.15, 11.x and 12.x.

Sadly it is difficult to reproduce and happens rarely. Here is a crash report: [R_2023-01-09-222950_Mac-1673302766269.txt](https://github.com/rstudio/httpuv/files/10377671/R_2023-01-09-222950_Mac-1673302766269.txt) taken from [this CI run](https://github.com/jeroen/curl/actions/runs/3877865517). The crash always happens in the same place, in the httpuv background thread:

```
Thread 1 Crashed:
0   libsystem_kernel.dylib        	0x00007fff2044991e __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff204785bd pthread_kill + 263
2   libsystem_c.dylib             	0x00007fff203cd406 abort + 125
3   httpuv.so                     	0x00000001092b9164 uv_mutex_unlock + 20 (thread.c:352)
4   httpuv.so                     	0x000000010925dd29 guard::~guard() + 8 (thread.h:38) [inlined]
5   httpuv.so                     	0x000000010925dd29 guard::~guard() + 8 (thread.h:37) [inlined]
6   httpuv.so                     	0x000000010925dd29 Barrier::wait() + 105 (thread.h:125)
7   httpuv.so                     	0x000000010926c145 io_thread(void*) + 149 (httpuv.cpp:126)
8   libsystem_pthread.dylib       	0x00007fff204788fc _pthread_start + 224
9   libsystem_pthread.dylib       	0x00007fff20474443 thread_start + 15
```

The main thread is pointing at some MacOS specific code:

```
2   httpuv.so                     	0x00000001092b927a uv_cond_destroy + 90 (thread.c:754)
```

This code apparently is already a hack for a macos threading bug:

https://github.com/rstudio/httpuv/blob/a1211dff0baf7250807048cfa216281f83b31731/src/libuv/src/unix/thread.c#L735-L762

Did something change recently in httpuv that could trigger this? I will try to investigate if I can reproduce the problem with older versions of httpuv.

wch:
Closes #351. This eliminates the race condition between the `Barrier` destructor and a call to `Barrier.wait()`.

Test code:

```
for i in {1..16}; do
    yes > /dev/null &
done

for i in {1..50}; do
    Rscript -e "s <- httpuv::startServer('0.0.0.0', 8000, list()); cat('.')"
done

killall yes
```

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -20,4 +20,4 @@ jobs:
   R-CMD-check:
     uses: rstudio/shiny-workflows/.github/workflows/R-CMD-check.yaml@v1
     with:
-      ubuntu: "ubuntu-18.04 ubuntu-20.04"
+      ubuntu: "ubuntu-20.04 ubuntu-latest"
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Package: httpuv
 Type: Package
 Encoding: UTF-8
 Title: HTTP and WebSocket Server Library
-Version: 1.6.7.9000
+Version: 1.6.7.9001
 Authors@R: c(
     person("Joe", "Cheng", role = c("aut"), email = "joe@rstudio.com"),
     person("Winston", "Chang", role = c("aut", "cre"), email = "winston@rstudio.com"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -38,7 +38,7 @@ Imports:
 LinkingTo: Rcpp, later
 URL: https://github.com/rstudio/httpuv
 SystemRequirements: GNU make, C++11, zlib
-RoxygenNote: 7.2.2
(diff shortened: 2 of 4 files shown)


jeroen:
Looks good: https://github.com/jeroen/httpuvtest/actions/runs/3891061945/jobs/6641550474

jcheng5 (approved):
It'd be great if you could change the top of `io_thread` to create a regular `std::shared_ptr<Barrier>` as early as possible and delete the `std::shared_ptr<Barrier>*`

PR #359, opened 2023-04-19
Send WS ping frames every 20 seconds

jcheng5:
- [x] Add NEWS
- [x] Manual testing
- [ ] Automated test?

## Manual testing performed

I used this build of httpuv with Shiny, both locally and deployed over Hugging Face Spaces, using Wireshark to make sure that Ping packets were being sent at the correct intervals.

I also used this torture test app:

```r
library(httpuv)

big_string <- strrep("j", 1e6)

app <- list(
  onWSOpen = function(conn) {
    stopped <- FALSE
    conn$onClose(function() {
      stopped <<- TRUE
    })


    cb <- function(...) {
      if (stopped) {
        return()
      }
      message("sending")
      conn$send(big_string)
    }
    cb()

    conn$onMessage(cb)
  }
)

httpuv::runServer("0.0.0.0", 8101, app)
```

In the browser on another machine, I navigated to `about:blank` and typed this code into the JS console:
```javascript
var ws = new WebSocket("ws://<HOSTNAME>:8101");
ws.onmessage = (msg) => {
  if (!/^j+$/.test(msg.data)) {
    console.log("bad");
  } else {
    console.log("good");
  }
  ws.send("");
};
```
What this does is repeatedly send large (1MB) messages from server to client, way too big to fit in a single TCP packet. I wanted to ensure that the Ping packets being sent at random times wouldn't get in between packets belonging to a message. The test worked; even hacking the httpuv source code to send Ping every 10 milliseconds didn't pose a problem.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # httpuv 1.6.9.9000
 
-* WebSocket connections now send Ping frames to the client every 20 seconds. This is only intended to serve as a keepalive for proxies that might be sitting in front of us; we don't pay attention to whether a Pong response is received in a timely manner.
+* WebSocket connections now send Ping frames to the client every 20 seconds. This is only intended to serve as a keepalive for proxies that might be sitting in front of us; we don't pay attention to whether a Pong response is received in a timely manner. (#359)
 
 
 # httpuv 1.6.9
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httpuv 1.6.9.9000
 
+* WebSocket connections now send Ping frames to the client every 20 seconds. This is only intended to serve as a keepalive for proxies that might be sitting in front of us; we don't pay attention to whether a Pong response is received in a timely manner.
+
 
 # httpuv 1.6.9
 
--- a/src/websockets.cpp
+++ b/src/websockets.cpp
@@ -220,6 +220,10 @@ void WSHyBiParser::read(const char* data, size_t len) {
   }
 }
 
+void WebSocketConnection::startPingTimer() {
+  uv_timer_start(&_pingTimer, pingTimerCallback, 20000, 20000);
+}
+
 bool WebSocketConnection::accept(const RequestHeaders& requestHeaders,
                                  const char* pData, size_t len) {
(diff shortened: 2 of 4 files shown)


src/websockets.cpp
```diff
@@ -274,6 +280,13 @@ void WebSocketConnection::sendWSMessage(Opcode opcode, const char* pData, size_t
                            safe_vec_addr(footer), footer.size());
 }
 
+void WebSocketConnection::sendPing() {
+  ASSERT_BACKGROUND_THREAD()
+  assert(_pParser);
```
wch, line 285: This won't do anything, right? Unless compiled with `-UNDEBUG`.

jcheng5: Yep, it's a true assert. I was getting a segfault cause `_pParser` was `NULL` and sending the ping frame uses `_pParser`. (I'm still not sure what caused that, I'll poke around before merging this)

jcheng5: Mystery solved (about why `_pParser` was null). The `WebSocketConnection` is created as soon as a socket connection is made; it's eagerly created, just in case we eventually need it. Because we speak HTTP/1.1, each socket can be kept alive and service many regular (non-WS) HTTP requests over time. In this case, Chrome was holding non-WebSocket-speaking connections open, and after 20 seconds, we tried to send a ping.

The only question this raises for me now, is whether it's a problem worth addressing that we allocate an additional `uv_timer_t` (don't start it, just initialize it) for every socket connection. I'm inclined to say no, personally.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # httpuv 1.6.9.9000
 
+* WebSocket connections now send Ping frames to the client every 20 seconds. This is only intended to serve as a keepalive for proxies that might be sitting in front of us; we don't pay attention to whether a Pong response is received in a timely manner.
```
wch, line 3:
```suggestion
* WebSocket connections now send Ping frames to the client every 20 seconds. This is only intended to serve as a keepalive for proxies that might be sitting in front of us; we don't pay attention to whether a Pong response is received in a timely manner. (#359)
```

wch (approved):

PR #368, opened 2023-05-10
Use system copy of libuv if found

jcheng5:
Adds a configure script that looks for libuv on the system that is version 1.43 (what we bundle) or newer. On macOS, static linking is preferred. If the library isn't found, installation falls backed to the bundled copy, as before. There is a new environment variable, `USE_BUNDLED_LIBUV`, that can be set to override this behavior: `true` means it won't look for system libuv, `false` means it won't build the bundled library and will fail if the system library isn't found.

Homebrew and Ubuntu 22.04 have new enough versions of libuv that could be installed. Our regular CI picks up the brew package in the macos-latest job, and I added a new workflow that tests the ubuntu package and the various combinations of the `USE_BUNDLED_LIBUV` variable.

### DIFF
--- a/.github/workflows/system-libuv.yaml
+++ b/.github/workflows/system-libuv.yaml
@@ -1,0 +1,79 @@
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, rc-**]
+  pull_request:
+  schedule:
+    - cron: '0 6 * * 1' # every monday
+
+name: Test system libuv
+
+jobs:
+  R-CMD-check:
+    # Ubuntu 22.04 has libuv1-dev version 1.43, which is sufficient
+    runs-on: ubuntu-latest
+    name: Installed=${{ matrix.config.install }} Bundled=${{ matrix.config.use }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          # This one should test that we've used system and everything runs ok
+          - {install: true}
+          # This one should test for bundling log message
+          - {install: true, use: true}
+          # This one should test for failure
+          - {install: false, use: false}
+
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      R_KEEP_PKG_SOURCE: yes
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::rcmdcheck
+          needs: check
+
+      - name: Install libuv
+        if: ${{ matrix.config.install }}
+        # If we add tests for e.g. rhel/centos, we couldn't assume apt-get
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libuv1-dev
+
+      - name: Ensure libuv is not on the system
+        if: ${{ !matrix.config.install }}
+        run: |
+          sudo apt-get remove -y libuv1-dev || true
+                   
+      - uses: r-lib/actions/check-r-package@v2
+        if: ${{ matrix.config.install }}
+        env:
+          USE_BUNDLED_LIBUV: ${{ matrix.config.use }}
+        with:
+          upload-snapshots: true
+      
+      - name: Confirm that USE_BUNDLED_LIBUV was respected
+        if: ${{ matrix.config.install }}
+        env:
+          USE_BUNDLED_LIBUV: ${{ matrix.config.use }}
+        run: |
+          if [ "$USE_BUNDLED_LIBUV" = "true" ]; then
+            grep "Using bundled copy of libuv" check/httpuv.Rcheck/00install.out
+          else
+            grep "Using libuv found by pkg-config" check/httpuv.Rcheck/00install.out
+          fi
+
+      - name: Confirm error if system libuv missing and use=false
+        if: ${{ !matrix.config.install && !matrix.config.use }}
+        env:
+          USE_BUNDLED_LIBUV: ${{ matrix.config.use }}
+        run: |
+          R CMD INSTALL . > install.log || true
+          grep "Did not find suitable libuv on your system" install.log
--- a/configure
+++ b/configure
@@ -1,0 +1,82 @@
+# Anticonf (tm) script by Jeroen Ooms (2021)
+# This script will query 'pkg-config' for the required cflags and ldflags.
+# If pkg-config is unavailable or does not find the library, try setting
+# INCLUDE_DIR and LIB_DIR manually via e.g:
+# R CMD INSTALL --configure-vars='INCLUDE_DIR=/.../include LIB_DIR=/.../lib'
+
+# Library settings
+PKG_CONFIG_NAME="libuv"
+PKG_DEB_NAME="libuv1-dev"
+PKG_RPM_NAME="libuv-devel"
+# PKG_CSW_NAME="libsodium_dev"
+PKG_BREW_NAME="libuv"
+PKG_TEST_HEADER="<uv.h>"
+PKG_LIBS="-luv"
+PKG_CFLAGS=""
+DEPS=""
+
+# Use pkg-config if available
+if [ `command -v pkg-config` ]; then
+  PKGCONFIG_CFLAGS=`pkg-config --cflags --silence-errors ${PKG_CONFIG_NAME}`
+  PKGCONFIG_LIBS=`pkg-config --libs ${PKG_CONFIG_NAME}`
+fi
+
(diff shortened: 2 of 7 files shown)


nealrichardson:
The builds don't print the package installation output, so you can't tell whether system or bundled libuv are being picked up. I made https://github.com/rstudio/shiny-workflows/pull/17 to try to add that to the GHA workflow.

configure
```diff
@@ -0,0 +1,82 @@
...
+  echo "with httpuv. If you would prefer to use a system copy, try installing:"
+  echo " * deb: $PKG_DEB_NAME (Debian, Ubuntu, etc)"
+  echo " * rpm: $PKG_RPM_NAME (Fedora, EPEL)"
+  # echo " * csw: $PKG_CSW_NAME (Solaris)"
+  echo " * brew: $PKG_BREW_NAME (OSX)"
+  echo "If $PKG_CONFIG_NAME is already installed, check that 'pkg-config' is in your"
+  echo "PATH and PKG_CONFIG_PATH contains a $PKG_CONFIG_NAME.pc file. If pkg-config"
+  echo "is unavailable you can set INCLUDE_DIR and LIB_DIR manually via:"
+  echo "R CMD INSTALL --configure-vars='INCLUDE_DIR=... LIB_DIR=...'"
+  echo "-------------------------- [ERROR MESSAGE] ---------------------------"
```
jcheng5, line 67: Make this less scary.

nealrichardson:
@schloerke I'm happy to take this on if you'd like

schloerke:
@nealrichardson yes, please! If you need to make changes to rstudio/shiny-workflows, please feel free!

nealrichardson:
In the regular CI, the ubuntu images do not have libuv installed so they all still bundle. The macOS job does appear to have it installed by homebrew, and the configure script picks it up and uses it fine.

nealrichardson:
For now, I've set the minimum version to 1.43, which is what we vendored. Looking at https://github.com/rstudio/httpuv/commits/main/src/libuv/include/uv/version.h, there have been few updates, and only [one](https://github.com/rstudio/httpuv/pull/214) mentions a reason for the upgrade. @wch do you happen to recall if there were reasons (bugfixes, new features) that motivated the updates in the past?

I'll check what ubuntu and centos et al. have in their repositories and see if they work in the test suite, though that might not be enough confidence that they're good to use if they're older than our version.

wch:
@nealrichardson Typically we have updated httpuv not for any specific reason, but just to keep up to date with bug fixes.

jcheng5:
I learned two new things about this topic yesterday that I don't fully understand.

First, from [CRAN Policies](https://cran.r-project.org/web/packages/policies.html) (emphasis mine):

> Where a package wishes to make use of a library not written solely for the package, the package installation should first look to see if it is already installed and if so is of a suitable version. In case not, it is desirable to include the library sources in the package and compile them as part of package installation. If the sources are too large, it is acceptable to download them as part of installation, but do ensure that the download is of a fixed version rather than the latest. Only as a last resort and with the agreement of the CRAN team should a package download pre-compiled software.
>
> **On Windows and macOS static libraries must be used.** A separate document, [External Libraries for CRAN packages](https://cran.r-project.org/web/packages/external_libs.html), covers what external libraries are or could be made available.

As written, when the local libuv is found, it's using the dynamically linked version. To get the statically linked version, it would be `pkg-config libuv-static` instead. Does the CRAN policy require us to use the static version? (I tried it the day I created this PR and it didn't immediately work, but I don't remember what went wrong)

---

Second, following the link in the previous quote, you get this paragraph:

> The build system for Windows changed with 4.2.0 and only that is considered here (and only 64-bit Windows is now supported).
>
> Building of external libraries is done using MXE (https://mxe.cc/). That has an extensive set of packages (https://github.com/mxe/mxe/blob/master/src), and users can contribute others (https://mxe.cc/#creating-packages). For external libraries not already available, the best approach is to contribute to the MXE library. For advice, consult Tomas Kalibera.

I don't know much about this new build system, but it certainly seems like MXE is supposed to help us with libuv somehow on Windows? Maybe it's just saying that on the Windows CRAN build machines, the static version of this library will be available??

nealrichardson:
> I learned two new things about this topic yesterday that I don't fully understand.
> 
> First, from [CRAN Policies](https://cran.r-project.org/web/packages/policies.html) (emphasis mine):
> 
> > Where a package wishes to make use of a library not written solely for the package, the package installation should first look to see if it is already installed and if so is of a suitable version. In case not, it is desirable to include the library sources in the package and compile them as part of package installation. If the sources are too large, it is acceptable to download them as part of installation, but do ensure that the download is of a fixed version rather than the latest. Only as a last resort and with the agreement of the CRAN team should a package download pre-compiled software.
> > **On Windows and macOS static libraries must be used.** A separate document, [External Libraries for CRAN packages](https://cran.r-project.org/web/packages/external_libs.html), covers what external libraries are or could be made available.
> 
> As written, when the local libuv is found, it's using the dynamically linked version. To get the statically linked version, it would be `pkg-config libuv-static` instead. Does the CRAN policy require us to use the static version? (I tried it the day I created this PR and it didn't immediately work, but I don't remember what went wrong)

I believe it's `pkg-config --static libuv`. I'll add that for macOS.

There is no "recipe" for libuv currently on https://github.com/R-macos/recipes, so I don't think the CRAN builders will have it. But if/when it does exist, we can pick it up. I don't think getting libuv on there will be a requirement for resubmission, I would guess that BDR will only look for the configure script to see that we are looking for system packages. But if we have to go that route, we can, and I'm hoping it's simple enough, it doesn't look like libuv requires complex configuration.

> 
> Second, following the link in the previous quote, you get this paragraph:
> 
> > The build system for Windows changed with 4.2.0 and only that is considered here (and only 64-bit Windows is now supported).
> > Building of external libraries is done using MXE (https://mxe.cc/). That has an extensive set of packages (https://github.com/mxe/mxe/blob/master/src), and users can contribute others (https://mxe.cc/#creating-packages). For external libraries not already available, the best approach is to contribute to the MXE library. For advice, consult Tomas Kalibera.
> 
> I don't know much about this new build system, but it certainly seems like MXE is supposed to help us with libuv somehow on Windows? Maybe it's just saying that on the Windows CRAN build machines, the static version of this library will be available??

libuv is on MXE, but it's an [older version](https://github.com/mxe/mxe/blob/master/src/libuv.mk), and IIUC we have a few patches we apply to our copy that haven't been upstreamed (yet). So I think that if we're asked about this (I doubt we will but you never know), we say that the version there is not adequate, and we'll work to get it upgraded for the next time we submit httpuv.

nealrichardson:
@jcheng5 I think this is done if you want to take a look (I can't request your review since technically it is your PR).

jcheng5:
> I believe it's `pkg-config --static libuv`. I'll add that for macOS.

OK. I think you're probably right, but please take a look at this issue if you haven't seen it already: https://github.com/libuv/libuv/issues/2988

.github/workflows/system-libuv.yaml
```diff
@@ -0,0 +1,79 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          # This one should test that we've used system and everything runs ok
+          - {install: true}
+          # This one should test for bundling log message
+          - {install: true, use: true}
+          # This one should test for failure
+          - {install: false, use: false}
```
jcheng5, line 25: Note to self. This is all of the possible combinations and their expected results:

|  | Install | Use | expected result |
|--|--------|--------|--------|
|  1 | true | unset | Uses system |
|  2 | false | unset | Uses bundled |
|  3 | true | true | Uses bundled |
|  4 | false | true | Uses bundled |
|  5 | true | false | Uses system | 
|  6 | false | false | Installation fails |

(I think you did a good job of narrowing it down to 3 configs)

jcheng5:
LGTM! I'm glad you volunteered to take this on, I am not currently equipped for this level of GHA sophistication.

nealrichardson:
> > I believe it's `pkg-config --static libuv`. I'll add that for macOS.
> 
> 
> 
> OK. I think you're probably right, but please take a look at this issue if you haven't seen it already: https://github.com/libuv/libuv/issues/2988
> 
> 
> 
> Edit: Actually, with what they're describing in that issue, the build would fail and ours is passing, so I withdraw the concern.

I'll double-check this locally. Brew installs both libuv.a and libuv.dylib. If it's working correctly, I should be able to brew uninstall libuv after installing the R package, and the R package should still work.

In practice it should be irrelevant for the CRAN builders.

nealrichardson:
Turns out the macOS linker always picks the dynamic library, and while gcc's linker has an [option](https://stackoverflow.com/questions/809794/use-both-static-and-dynamically-linked-libraries-in-gcc/816138#816138) to control this, `ld` on macOS [does not](http://www.manpagez.com/man/1/ld64/). I'll come up with a workaround.

PR #377, opened 2023-05-22
Support websocket compression (permessage-deflate)

jcheng5:


zeehio:
On shiny websites with plotly plots with a large number of points, generating the plot is the first bottleneck. Once that is cached (bindCache...) the transfer of the plotly object from server to client becomes the bottleneck.

Having websocket per-message compression would reduce the cost of the transfer at the expense of compressing/decompressing each message.

Ideally, having some websocket-based caching system would be awesome, so the web browser does not download the same plot more than once. I would imagine httpuv sending a hash of the response first, and then, if the browser does not have anything matching that hash, the full response (maybe with compression as implemented here).

Thanks!

anthonynorth:
Looking for permessage-deflate support in shiny led me here. Are there plans to merge this into main?

TBlackmore:
Also came here looking to speed up shiny messages. Can I upvote this feature?

PR #378, opened 2023-07-21
Remove httpuv_decodeURIComponent workaround

nealrichardson:
Just noticed this "we can probably remove this in a year" comment... from 2017.


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# httpuv 1.6.11.9000
+
+* Remove a workaround to support `shiny` older than version 1.0.6 (2017)
```
wch, line 3:
```suggestion
* Remove a workaround to support `shiny` older than version 1.0.6 (#378)
```

wch (approved):
Looks good, thanks!

NEWS.md
nealrichardson: FTR I meant this as a year, not PR number, but this is fine

wch: Ah, that makes sense.

PR #410, opened 2025-04-14
Fix libuv R CMD check warnings

shikokuchuo:
1. Fix follows the same strategy as @gaborcsardi in 'fs'
2. In configure - simplifies static linking in MacOS by passing the path to the .a file directly rather than try to copy it
3. Updates logic in Makevars.ucrt for when pkg-config doesn't work (causing the CI failure seen on R4.2) - based on @jeroen's patch to 'websocket'


wch:
The .a thing is kind of nutty, but hey, if it works...

Thanks!

PR #427, opened 2025-08-12
perf: Use native symbol registration

### LINKED ISSUE #426: Use native symbol registration
Came across this from another codebase.

This is as close as you get to a 'free lunch'. Using `.registration = TRUE` speeds up all `.Call()` calls into native code.

shikokuchuo:
Closes #426. Free speed up for all calls into native code.

From Writing R Extensions: https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Registering-native-routines

> This avoids the overhead of looking up an entry point each time it is used, and ensures that the entry point in the package is the one used (without a PACKAGE = "pkg" argument).

Similar changes in https://github.com/r-lib/later/pull/187 and https://github.com/mlverse/torch/pull/976.


R/RcppExports.R
```diff
@@ -2,98 +2,98 @@
...
 #' browsers do. The precise behaviors of these functions can be found at
 #' developer.mozilla.org:
-#' [encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI),
-#' [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent),
-#' [decodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI),
-#' [decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent)
+#' \href{https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI}{encodeURI},
+#' \href{https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent}{encodeURIComponent},
+#' \href{https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI}{decodeURI},
+#' \href{https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent}{decodeURIComponent}
```
schloerke, line 56: These changes work, but https://github.com/rstudio/httpuv/blob/8f0b65d75119e70b87bf137684ebe77c052f7932/DESCRIPTION#L78 states markdown is the default format.

R/httpuv-package.R
```diff
@@ -22,7 +22,7 @@
 #' @title HTTP and WebSocket server
 #' @author Joe Cheng \email{joe@@rstudio.com}
 #' @keywords package
-#' @useDynLib httpuv
+#' @useDynLib httpuv, .registration = TRUE
```
schloerke, line 25: This is new to me (due to lack of opportunities). Can you like the related PR and/or docs?

R/RcppExports.R
schloerke: Similar changes in other doc changes could be reverted.

shikokuchuo: I'll take another look from my Linux machine. This is autogenerated by `Rcpp::compileAttributes()` so not sure why it no longer respects markdown.

R/httpuv-package.R
shikokuchuo: Added to the main PR comments above.

R/RcppExports.R
schloerke: Ah. If it's auto generated, then I'm fine with the changes.

I had manually converted all files myself.

Is `Rcpp::compileAttributes()` something that we should add to our routine checks if Rcpp is used?

R/httpuv-package.R
wch: @schloerke The relevant docs are here:

- https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#useDynLib
- https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Registering-native-routines

From the second one:
> This avoids the overhead of looking up an entry point each time it is used

I recall doing some speed tests with `.registration = TRUE` a while back, and it was a bit faster than not using it.

R/RcppExports.R
shikokuchuo: Manually 😓 ?! It's not normally necessary, but I think I had to manually run it for this particular change.

schloerke: No changes needed. Resolving

schloerke (approved):
Thank you!

schloerke:
Please add a NEWS entry before merging

R/RcppExports.R
shikokuchuo: I see now that as it was manually modified, it was not automatically updated by `devtools::document()` or the equivalent, and why I had to call `Rcpp::compileAttributes()` manually. Now the file will follow any changes as needed and we don't need to worry about it being updated.

Issue #429, opened 2025-11-25
httpuv fails to build on macOS if automake is present


s-u:
If `automake` is present (i.e., on the `PATH`) then `httpuv` fails to build because it tries to re-generate `configure` in a fundamentally broken way, thus failing the installation.

Expected: it should not try to regenerate without maintainer mode being enabled and thus regular installation should work regardless of the absence/presence of `automake`.

To replicate, install `automake` and its dependenceis (here `automake-1.17` from  https://mac.r-project.org/bin/ - but it breaks even if you try to match the internally used versions). The problem does not occur is `automake` is not on the `PATH`.

```
* installing *source* package ‘httpuv’ ...
** this is package ‘httpuv’ version ‘1.6.16’
** package ‘httpuv’ successfully unpacked and MD5 sums checked
** using staged installation
** libs
using C compiler: ‘Apple clang version 17.0.0 (clang-1700.3.19.1)’
using C++ compiler: ‘Apple clang version 17.0.0 (clang-1700.3.19.1)’
using SDK: ‘MacOSX14.5.sdk’
aclocal.m4:17: warning: this file was generated for autoconf 2.71.
You have another version of autoconf.  It may work, but is not guaranteed to.
If you have problems, you may need to regenerate the build system entirely.
To do so, use the procedure documented by the package, typically 'autoreconf'.
+ glibtoolize --copy
autogen.sh: line 43: glibtoolize: command not found
Error in untar2(tarfile, files, list, exdir, restore_times, allow_all) : 
  incomplete block on file
rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory
/Volumes/Temp/tmp/Rtmpc9VX5f/R.INSTALL4d7147fa73de/httpuv/src/libuv/missing: line 81: aclocal-1.16: command not found
WARNING: 'aclocal-1.16' is missing on your system.
         You should only need it if you modified 'acinclude.m4' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'aclocal' program is part of the GNU Automake package:
         <https://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <https://www.gnu.org/software/autoconf>
         <https://www.gnu.org/software/m4/>
         <https://www.perl.org/>
make[1]: *** [aclocal.m4] Error 127
[...]

Issue #341, opened 2022-05-06
Installation failure with R 4.2.0


Deleetdk:
I am on Mint 19.3 (Ubuntu 18.04 based) and just upgraded to R 4.2.0:

```
> sessionInfo()
R version 4.2.0 (2022-04-22)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Linux Mint 19.3

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1
```

I cannot install this package again. The CRAN and Github versions give the same error:

```
> devtools::install_github("rstudio/httpuv")
Downloading GitHub repo rstudio/httpuv@HEAD
✔  checking for file ‘/tmp/RtmpThQuzi/remotes74c81dee06d5/rstudio-httpuv-220319d/DESCRIPTION’ ...
─  preparing ‘httpuv’:
✔  checking DESCRIPTION meta-information ...
─  cleaning src
─  checking for LF line-endings in source and make files and shell scripts
─  checking for empty or unneeded directories
─  building ‘httpuv_1.6.5.9000.tar.gz’
   
Installing package into ‘/home/emil/R/x86_64-pc-linux-gnu-library/4.2’
(as ‘lib’ is unspecified)
* installing *source* package ‘httpuv’ ...
** using staged installation
** libs
g++ -std=gnu++11 -I"/usr/share/R/include" -DNDEBUG -Ilibuv/include -pthread -I'/home/emil/R/x86_64-pc-linux-gnu-library/4.2/Rcpp/include' -I'/home/emil/R/x86_64-pc-linux-gnu-library/4.2/later/include'   -fvisibility=hidden -DSTRICT_R_HEADERS -fpic  -g -O2 -fdebug-prefix-map=/build/r-base-kt0bjq/r-base-4.2.0=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2  -c RcppExports-legacy.cpp -o RcppExports-legacy.o
[...]

Issue #338, opened 2022-04-20
automake: Perl lib version (5.28.2) doesn't match executable


xiucz:
Hi,

When I want to install the package,  it shows the following error:

> R version 4.1

```
CC="gcc -std=gnu99" CFLAGS="-g -O2  -fpic -fvisibility=hidden" AR="ar" RANLIB="ranlib" LDFLAGS="-L/usr/local/lib64" ./configure "--quiet"
automake found. Running autogen.sh.
/usr/bin/automake: Perl lib version (5.28.2) doesn't match executable '/usr/bin/perl' version (5.16.3) at ~/PERL/perl-5.28.2/lib/Config.pm line 62.
Compilation failed in require at ~/PERL/perl-5.28.2/lib/threads.pm line 13.
Compilation failed in require at /usr/bin/automake line 138.
BEGIN failed--compilation aborted at /usr/bin/automake line 143.
autogen.sh: line 33: test: : integer expression expected
autogen.sh: line 34: test: : integer expression expected
+ libtoolize --copy
```
My perl version, I know it is related to the perl version ,  I use my own perl , not the systerm perl(/usr/bin/perl). 
```
$ which perl 
~/PERL/perl-5.28.2/bin/perl
````
But I dont know how to modify the code, can you give me some advice?

Thank you,
xiucz

**Comment by xiucz** [comment]
Solved by 
```

$ PERL5LIB=''
#then start R and install
$ R
```

Issue #331, opened 2022-01-05
httpuv now not compatible with Ubuntu 18? 


nick-youngblut:
Installation of httpuv on my system (Ubuntu 18.04.5; R 4.1.2) results in the following error:

```
g++ -std=gnu++11 -I"/opt/R/4.1.2/lib/R/include" -DNDEBUG -Ilibuv/include -pthread -I'/ebio/abt3_projects/software/dev/shiny-apps/RTecanFluentShiny/renv/staging/1/Rcpp/include' -I'/ebio/abt3_projects/software/dev/shiny-apps/RTecanFluentShiny/renv/staging/1/later/include' -I/usr/local/include  -fvisibility=hidden -DSTRICT_R_HEADERS -fpic  -g -O2  -c winutils.cpp -o winutils.o
cp -p -f libuv/m4/lt_obsolete.m4 libuv/m4/lt~obsolete.m4
cd libuv; \
if ! command -v automake >/dev/null 2>&1 ; then \
	echo "automake not found. Touching files so configure will not try to run automake."; \
	touch aclocal.m4; \
	touch -r aclocal.m4 configure Makefile.in; \
else \
	echo "automake found. Running autogen.sh."; \
	sh autogen.sh; \
fi; \
chmod +x configure; \
CC="gcc" CFLAGS="-g -O2  -fpic -fvisibility=hidden" AR="ar" RANLIB="ranlib" LDFLAGS="-L/usr/local/lib" ./configure "--quiet"
automake found. Running autogen.sh.
+ libtoolize --copy
autogen.sh: 43: autogen.sh: libtoolize: not found
make --directory=libuv \
	HAVE_DTRACE=0
make[1]: Entering directory '/tmp/RtmpJ9DCd6/renv-package-56c546d4b1f1/httpuv/src/libuv'
CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/bash /tmp/RtmpJ9DCd6/renv-package-56c546d4b1f1/httpuv/src/libuv/missing aclocal-1.16 -I m4
/tmp/RtmpJ9DCd6/renv-package-56c546d4b1f1/httpuv/src/libuv/missing: line 81: aclocal-1.16: command not found
WARNING: 'aclocal-1.16' is missing on your system.
         You should only need it if you modified 'acinclude.m4' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'aclocal' program is part of the GNU Automake package:
         <https://www.gnu.org/software/automake>
[...]

PR #430, opened 2026-03-09
fix: Stop running autogen.sh when building bundled libuv

cpsievert:
## Problem

httpuv is failing to install on [CRAN `r-devel-macos-arm64`](https://cran.r-project.org/web/checks/check_results_httpuv.html) with:

```
autogen.sh: line 43: glibtoolize: command not found
WARNING: 'aclocal-1.16' is missing on your system.
make: *** [libuv/.libs/libuv.b] Error 2
```

## Root cause

When building the bundled libuv, `src/Makevars.in` checks whether `automake` is on the system PATH:

- **No automake found** → touches timestamps on pre-generated autotools files so `make` doesn't try to regenerate them, then runs `./configure` (safe path)
- **automake found** → runs `autoupdate` + `autogen.sh` to regenerate everything from scratch, then runs `./configure`

On the CRAN `r-devel-macos-arm64` machine, `automake` is present (so the `else` branch fires), but the rest of the autotools toolchain (`glibtoolize`, matching `aclocal-1.16`) is missing or version-mismatched. `autogen.sh` calls these tools and fails.

This automake-detection logic was added in #319 (Dec 2021) to handle cases where a newer system automake caused timestamp-related regeneration issues (#280). However, the regeneration approach introduced its own fragility — it requires a complete, version-matched autotools toolchain.

## Fix

Three layers of defense, each sufficient on its own:

### 1. Remove `autogen.sh` at build time (1ec74b9)

Remove the `if/else` automake detection and `autoupdate`/`autogen.sh` calls from `Makevars.in`. The bundled libuv ships pre-generated autotools output — there is no reason to regenerate at build time. This eliminates all "Class 2" failures (#331, #338, #341, #429) where `autogen.sh` was called with an incomplete toolchain.

### 2. Touch inputs AND outputs (071c300)

The pre-#319 touch approach only touched **output** files (`aclocal.m4`, `configure`, `Makefile.in`). The generated `Makefile` contains regeneration rules that compare input timestamps (`Makefile.am`, `configure.ac`, `m4/*.m4`) against outputs. If any input has a newer timestamp (e.g. due to extraction ordering), `make` fires these rules — this was the cause of #124 and #280.

We now touch **all** autotools files, inputs first then outputs set to match, ensuring `mtime(inputs) <= mtime(outputs)`.

### 3. `AM_MAINTAINER_MODE` (913474c)

Added `AM_MAINTAINER_MODE` to libuv's `configure.ac` and regenerated the autotools files. This wraps all regeneration rules in `Makefile.in` behind a `@MAINTAINER_MODE_TRUE@` conditional that is disabled by default. Without `--enable-maintainer-mode` passed to `./configure`, the rules are structurally absent from the generated `Makefile` — `make` never sees them regardless of timestamps or installed toolchain.

This is the standard approach used by projects that ship pre-generated autotools output (curl, libffi, libexpat, libxml2).

## Historical evidence

Every libuv build failure in this repo's history falls into one of two classes:

### Class 1: Regeneration triggered despite intent to prevent it (#124, #280)

| Issue | Platform | Failure | Fixed by |
|-------|----------|---------|----------|
| #124 | CRAN Fedora | automake 1.15 vs committed 1.15.1 | Touch inputs (#2) + `AM_MAINTAINER_MODE` (#3) |
| #280 | Alpine 3.12 | automake 1.16.2 vs committed 1.16.1 | Touch inputs (#2) + `AM_MAINTAINER_MODE` (#3) |

### Class 2: `autogen.sh` called with incomplete toolchain (#331, #338, #341, #429)

| Issue | Platform | Missing tool | Fixed by |
|-------|----------|-------------|----------|
| #331 | Ubuntu 18 | libtoolize | Remove autogen.sh (#1) |
| #338 | User system | Perl version mismatch in automake | Remove autogen.sh (#1) |
| #341 | Ubuntu 18 | libtoolize | Remove autogen.sh (#1) |
| #429 | CRAN macOS arm64 | glibtoolize + aclocal-1.16 | Remove autogen.sh (#1) |


shikokuchuo:
@copilot+gpt-5.4 review

shikokuchuo (changes requested):
Removing the conditional branch would seem to re-introduce https://github.com/rstudio/httpuv/issues/280, which would trigger when the system automake version is different to that used to generate the committed files?

cpsievert:
Thanks for catching that @shikokuchuo! [071c300](https://github.com/rstudio/httpuv/pull/430/commits/071c30068a3d63b47839358c05861b69c868663f) should address the issues there. Also note the updated PR description with the reasoning behind It addressing #280. 

I also had Claude do a bit of research into other issues that have reported install problems. It generated a little summary that's also included in the PR description.

shikokuchuo:
Below solution suggested by Claude Code. I was sceptical about the diagnosis for "Class 1" problems that the "input files (Makefile.am, configure.ac, m4/*.m4) had newer timestamps than outputs after extraction". Given we just touched the output files, how could the input files have newer timestamps? => Removing the regeneration logic entirely seems to be surer.

---

Thanks for the fix — removing the `autogen.sh` path is clearly the right call and fixes #429 and the whole "class 2" (incomplete toolchain) category.

One suggestion for a more robust solution to the "class 1" failures (#124, #280) where `make` fires regeneration rules despite the touch. The touch approach works in practice, but the proper autotools mechanism for this is `AM_MAINTAINER_MODE`. Adding it to `configure.ac`:

```
AM_INIT_AUTOMAKE([-Wall -Werror foreign subdir-objects] UV_EXTRA_AUTOMAKE_FLAGS)
AM_MAINTAINER_MODE
```

This causes automake to wrap all regeneration rules in the generated `Makefile.in` behind a `@MAINTAINER_MODE_TRUE@` conditional, which is disabled by default. Without `--enable-maintainer-mode` passed to `./configure`, the rules are completely absent from the generated `Makefile` — `make` never sees them, so it can never try to run automake/autoconf regardless of timestamps.

This is the standard approach used by projects that ship pre-generated autotools output (curl, libffi, libexpat, libxml2 all use it).

With `AM_MAINTAINER_MODE` in place, the touch logic could be dropped entirely, or kept as belt-and-suspenders. Either way, the regeneration rules would be structurally eliminated rather than suppressed via timestamp manipulation.

This would need to be done at "Phase 1" time (developer runs `autogen.sh` and commits the regenerated files).

shikokuchuo (approved):
Great! I would not still `touch` the files, but it shouldn't do any harm.

PR #432, opened 2026-03-10
fix: Skip tests if suggested packages aren't available

cpsievert:
## Summary

- `curl` is in `Suggests`, but `library(curl)` was called unconditionally in `tests/testthat/helper-app.R`, causing the entire test suite to fail when curl is not available (e.g. in the `check-depends-only` CI job)
- The helper now loads curl conditionally with `requireNamespace()`
- The 4 test files that depend on curl (`test-app`, `test-http-parse`, `test-static-paths`, `test-static-server`) skip when curl is not installed
- The remaining 3 test files (`test-utils`, `test-traffic`, `test-frame-completion`) run regardless

## Test plan

- [ ] CI `check-depends-only` job passes
- [ ] All other CI jobs maintain the same pass/skip/fail counts

### DIFF
--- a/tests/testthat/helper-app.R
+++ b/tests/testthat/helper-app.R
@@ -1,6 +1,3 @@
-if (requireNamespace("curl", quietly = TRUE)) {
-  library(curl)
-}
 library(promises)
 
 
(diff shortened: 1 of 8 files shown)


tests/testthat/helper-app.R
```diff
@@ -1,4 +1,6 @@
-library(curl)
+if (requireNamespace("curl", quietly = TRUE)) {
+  library(curl)
+}
```
cpsievert: I think I'd prefer using `curl::` where it's actually relevant.

PR #438, opened 2026-05-14
Refactor codebase to use 'cpp11'

pachadotdev:
Dear 'httpuv' maintainers:

I hope you are doing well.

I took all the feedback from the previous thread, and decided to try with cpp11.

I registed functions of the form `my_fun_()` on C++ side and roxygenised wrappers of the form `my_fun()` on R side, the changes go smoothly.

Similar to 'later' (PR [269](https://github.com/r-lib/later/pull/269)), but without a side 'httpuvtest' package as there is no API here, I looped multiple `devtools::check()` calls for different pairs such as "Clang + CXX17/20/23" and "GCC + CXX17/20/23". In addition, same for 'later', I use https://github.com/wch/r-source to run `make build-r-devel && make check-devel`  to build R-devel in a sub-directory and there install devtools and all the dependencies to run an extra `devtools::check()`.

I know this testing may read a circular but Clang usually shows things that GCC accepts silently.

Please let me know about this PR. I tried to edit as little as possible.

Be safe and well,
MVS


pachadotdev:
@estebanlp here's the other, same ABI thing

